PR #1, opened 2026-04-28
Add support for reverse proxy, fix all vulnerabilities, add missing "dev" deps, 

0x3dlux:
- **Reverse proxy / WSS**: Updated `ws.ts` to auto-detect `wss://` vs `ws://` and use the correct port (`443`/`80` for proxied, `10469` for local dev) with the option to have the web server wrap everything in Basic auth
- **Local vite overrides**: Added support for `vite.config.local.ts` (gitignored) via `mergeConfig`, so host-specific settings like `allowedHosts` don't pollute the repo
- **Missing deps**: Added `@types/express`, `@types/cors`, `serialize-javascript`, and `vite-plugin-pwa` to `devDependencies`
- **Cleanup**: Removed unused `.env.example` (dotenv is never actually loaded by the app)
- **Docs**: Added `REVERSE-PROXY.md` with nginx, HAProxy, and Traefik examples, plus `vite.config.local.ts.example`

PS: Fellow GX10'er here. Found your project through the DGX Spark forum and instantly recognized it as the better OpenCode. Commits have been produced by Qwen3.6-27B (M2.7 doesn't run on a single Spark :-/) via OpenCode which is now quickly being replaced by OpenFox. Hats off to what you did here in not even 2 months! I actually had a similar idea to rewrite OpenCode in Rust around a month ago but: E_TIME, as usual...


0x3dlux:
That huge mess in package-lock.json leads to this:

image

co-l:
Hey! Sorry I didn't notice your work earlier, I'm not used to have contributors on GitHub haha.

I'm not sure on how to move forward, we should probably have a call or exchange messages first.

0x3dlux:
Sure. PN...

PR #3, opened 2026-04-30
fix: replace intrusive Ctrl/Shift double-press with Ctrl+J and Ctrl+K

0x3dlux:
## Summary
  
- Replace double-press Ctrl (terminal) and Shift (quick action) with standard **Ctrl+J** and **Ctrl+K** shortcuts
- On Windows, holding Ctrl/Shift to type frequently causes these handlers to misfire, especially with sticky keys enabled
- ESC to abort generation is now gated: it won't fire when any popup is open (quick action, commands, workflows, message search), preventing accidental aborts when closing a modal
- ESC to close terminal (Ctrl+J) works as before — it's in a separate handler in TerminalDrawer

### DIFF
--- a/web/src/components/layout/Header.tsx
+++ b/web/src/components/layout/Header.tsx
@@ -326,20 +326,13 @@ export function Header({ onMenuClick, onCriteriaToggle }: HeaderProps) {
 
   
 
-  // Double Ctrl opens terminal
-  const lastCtrlRef = useRef<number>(0)
+  // Ctrl+J opens terminal
   useEffect(() => {
     const handleKeyDown = (e: KeyboardEvent) => {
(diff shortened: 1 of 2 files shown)


co-l:
Hey for this one, it'll break my muscle memory.

I'm ok with making this the default though, but we'll need to add a way for customizing behaviour.

0x3dlux:
Haha, fine with me. Configurable hotkeys seems like the ideal solution here indeed. I didn't mean to impose any random key bindings, and was just scratching my own itch. For me the quick action menu pretty much always jumped at me when I typed in a capital letter. My keyboard's repeat delay being set to minimum may be to blame here, so I'm basically allowed to keep Shift down for about 100ms before that registers as "double shift".

co-l:
Sorry for taking so long, but this is now possible through the Settings modal:

image

PR #5, opened 2026-05-01
feat: show command text in UI before execution

0x3dlux:
Pipe partial tool arguments through the tool.preparing event so the UI can display the actual command text as it streams in from the LLM, instead of only showing "Running command...".


0x3dlux:
image
while it's still "blinking"

PR #6, opened 2026-05-01
feat: show project path in project settings dialog

0x3dlux:
Show the project's working directory path in the Edit project settings modal. Adds an infoRows prop to InstructionsModal for extensible static info fields.


0x3dlux:
image

PR #8, opened 2026-05-04
fix: remove gap between Send and MoreMenu buttons

0x3dlux:
Make the Send and three-dots (MoreMenu) sit flush against each other again. Was introduced unintentionally in my feature/stop-button branch.

### DIFF
--- a/web/src/components/plan/PlanPanel.tsx
+++ b/web/src/components/plan/PlanPanel.tsx
@@ -701,6 +701,7 @@ export function PlanPanel({ criteriaSidebarOpen: externalCriteriaSidebarOpen, on
               Abort
             </button>
           )}
+                <div>
                 <button
                   type="button"
                   onClick={() => {
@@ -745,6 +746,7 @@ export function PlanPanel({ criteriaSidebarOpen: externalCriteriaSidebarOpen, on
                   criteria={session?.criteria ?? []}
                 />
               </div>
+                </div>
             </div>
         <div className="mt-3 flex items-center justify-between">
           <div className="flex items-center gap-2">
--- a/web/src/components/plan/PlanPanel.tsx
+++ b/web/src/components/plan/PlanPanel.tsx
@@ -701,7 +701,7 @@ export function PlanPanel({ criteriaSidebarOpen: externalCriteriaSidebarOpen, on
               Abort
             </button>
           )}
-                <div>
+                <div className="flex items-center">
                 <button
                   type="button"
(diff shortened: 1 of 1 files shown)


0x3dlux:
Should be clean now.

co-l:
Hey! I get notifications now haha. Yeah I noticed it as well and was going to fix it.

For the prettier normalization, I've made the process more streamlined with precommit hooks that run format + verification (typecheck, lint, test) so we "can't" commit broken state. (always doable with --no-verify)

I also have a smoke test on `npm publish` to check if basic flow works (onboarding, adding a project, starting a session, launching a test workflow).

I was comfortable not having an CI, but maybe I could set that up at some point.

0x3dlux:
Great :) The linter is indeed a bit annoying right now. Qwen keeps working around it by using sed instead of it's edit tool and always using --no-verify to keep PRs nice and clean. The pre-commit hook only reformats the files that have been touched in that commit. My reverse proxy + **npm audit** PR may have upgraded the prettifiers to be more strict (?) so I guess it would make sense to do a manual re-lint of the entire codebase in a single commit so the hooks won't stand in the way of nice and clean commits anymore.

co-l:
Ok I'm running something like a global pass right now, you'll tell me if you encounter any annoyance from link/format in the future.

I was so tired of agents using --no-verify that I added a regex check for requiring user validation (it's a bit buggy now as it flags anything with "-n" and it can't be disabled by default). Goal is having the agent always honour the feedback from the hooks instead of blindly bypassing them.

> Qwen keeps working around it by using sed instead of it's edit tool

That's weird, is that while using OpenFox or OpenCode?

0x3dlux:
I still use OpenCode to work on OpenFox, to not land myself in a chicken-egg situation, but all my other projects live in OpenFox by now. Qwen in OpenCode identified sed as a workaround to not trigger what it believes to be Vite's HMR doing the reformatting - but that's actually not the case, right?

co-l:
Yeah the chicken egg situation is real, since I'm working solely on OpenFox WITH OpenFox I get what you mean :D but I've made it possible, it's my main use case right now so I've made it work.

There shouldn't be any live reformatting ; it should only occur with precommit hooks. So it's still weird.

0x3dlux:
Pretty sure there was no live reformatting going on, Qwen just mis-identified the pre-commit's actions as Vite's HMR (and kept on killing the dev server which solved nothing).

PR #13, opened 2026-05-19
feat: add conversation history to the sidebar

0x3dlux:
- Add conversation history sidebar with bi-directional scroll synchronization
- Display conversation timeline with icons for different message types
- Prioritize regular content over thinking content (for items containing both)
- Show workflow and task completion status with proper formatting
- Emoji priority order for the sidebar (from highest to lowest priority):

⚡ — Workflow Started (messageKind === 'workflow-started')
✅ — Task Completed (messageKind === 'task-completed' or summary in JSON)
📝 — Auto-Prompt (messageKind === 'auto-prompt')
🔧 — Correction (messageKind === 'correction')
🔄 — Context Reset (messageKind === 'context-reset')
⚡ — Command Executed (messageKind === 'command')
🔨 — Build Mode (System reminder with # Build Mode)
📋 — Plan Mode (System reminder with # Plan Mode)
🗜️ — Compaction (Content includes "compacted", "compaction", or "context window")
⚙️ — System Reminder (Generic system message)
🧠 — Thinking (Assistant message with thinking content but NO regular content)
🤔 — User Message (Regular user message)
🤖 — Assistant Message (Regular assistant message with content)

image

I have smooth scrolling disabled in my Chrome, so I've captured Firefox here.

https://github.com/user-attachments/assets/93fb4281-addb-4d5b-bfa0-f15ae9c67d72

Starting from ConversationIndex.tsx:405 this is particularly "ugly" - if you want to call "explicit verbosity" that way, but it gets the job done. Sorry this has turned into a particularly big PR, but this feature definitely helps me get back on track quickly when hopping 42 projects at once ;)

I will handle Firefox's ugly scrollbars soon-ish too...

### DIFF
--- a/web/src/components/layout/SessionLayout.tsx
+++ b/web/src/components/layout/SessionLayout.tsx
@@ -2,12 +2,16 @@ import type { ReactNode } from 'react'
 import { useSessionStore } from '../../stores/session'
 import { SummaryDisplay } from '../plan/SummaryDisplay'
 import type { Message } from '@shared/types.js'
+import type { DisplayItem } from '../plan/groupMessages'
 
 interface SessionLayoutProps {
   children: ReactNode
   criteriaSidebarOpen?: boolean
   onCriteriaSidebarToggle?: () => void
   messages: Message[]
+  displayItems?: DisplayItem[]
+  activeIndex?: number
+  onNavigate?: (index: number) => void
 }
 
 export function SessionLayout({
@@ -15,6 +19,9 @@ export function SessionLayout({
   criteriaSidebarOpen = true,
   onCriteriaSidebarToggle,
   messages,
+  displayItems,
+  activeIndex,
+  onNavigate,
 }: SessionLayoutProps) {
   const session = useSessionStore((state) => state.currentSession)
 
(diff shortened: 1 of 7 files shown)


0x3dlux:
One single remaing issue: the items at the very bottom are kind of unreachable via scrolling alone, so things act a bit weird there. I somewhat worked around this by highlighting the bottom item whenever we're "Live" (scrolled to the very bottom) and "faking" the highlight in the index when they're clicked, but scrolling the chat window afterwards will again "misplace" the highlighted item. We're kind of trying to work against "physics" here in a sense though, so I think this is an "ok" compromise...

0x3dlux:
Sorry, mis-deleted

0x3dlux:
btw: what's your plan with the "Summary" section? I never saw it show anything other than "No summary yet", and Qwen is also just throwing a bunch of guesses:

----
**The Problem**
The session summary is supposed to be generated by the planner agent when it finishes its planning phase, but there's no code that actually captures the planner's output and saves it to the session.

**What should happen:**
Planner mode runs and defines criteria
Planner should output a summary of the plan
That summary should be saved to session.summary via sessionManager.setSummary()
The summary would then be displayed in the right sidebar and used by the verifier

**What actually happens:**
The session.summary field is never populated during planner mode
The summary is only set in one place: SessionManager.setSummary() at line 329 of src/server/session/manager.ts
This method is never called during normal planner → builder workflow transitions

**Evidence:**
* Only 2 references to setSummary exist:
  * The method definition itself
  * One test case
* No workflow or agent code calls it to save the planner's output
* The planner agent definition (planner.agent.md) doesn't instruct the planner to output a summary via return_value

**Result:**
The summary field stays null forever, so you always see "No summary yet" in the sidebar. The verifier workflow template at line 34 uses {{summary}} but it's always empty.

This is a missing feature/bug - the planner workflow needs to be updated to capture and save its summary output when transitioning to builder mode.

----

Isn't most of that handled by in-line "Acceptance criteria" by now?

0x3dlux:
@co-l, you're still getting notifications, right?

co-l:
Hey sorry I've been really busy the past two weeks, didn't have much time (I was giving an AI dev training and it consumed 100% of my energy).

I'll merge your PR next week after I go back home

co-l:
Hey!

About the summary: I think I'm going to remove it at some point.

I've taken your feature and integrated it into the existing message history modal (and made it easier to find:)
image

image

It's live on main and in 1.6.95

PR #15, opened 2026-05-19
style: Add Firefox-compatible thin scrollbars

0x3dlux:
## Changes
- Added Firefox scrollbar styling using standard CSS properties
- Used thin scrollbar width with subtle semi-transparent colors
- Scrollbars now match Chrome's appearance across both browsers
- Transparent track and 30% opacity thumb for minimal visual intrusion

## Testing
- Verified scrollbars appear consistent in both Chrome and Firefox

### DIFF
--- a/web/src/styles/globals.css
+++ b/web/src/styles/globals.css
@@ -32,6 +32,12 @@
   ::-webkit-scrollbar-thumb:hover {
     @apply bg-text-muted;
   }
+
+  /* Firefox scrollbars */
+  * {
+    scrollbar-width: thin;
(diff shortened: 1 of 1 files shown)


co-l:
Hey! Not sure about this one, behavior of scrollbar in Firefox seems correct to me right now, out of the way when not needed and grows on hover.

0x3dlux:
Looks like this is indeed a Windows-only issue. The hardcoded RGB value was kind of off too, so I gave this another run.

Before:
image

After:
imageimage

PR #16, opened 2026-05-19
feat: add "Show syntax highlighting" setting

0x3dlux:
- Add DISPLAY_SHOW_SYNTAX_HIGHLIGHTING setting key to server and client
- Default to enabled for backward compatibility
- Add toggle in Settings > Display below 'Show workflow bars'
- When disabled, render code blocks with basic monospace styling
- Affects both CodeHighlight component and Markdown code blocks

SyntaxHighlighter3 is **!!!RIDICULOUSLY!!! inefficient**!
Switching to a session with 47 messages and 8 code blocks and syntax highlighting enabled takes a whopping **3.65s**

Highlighter enabled: 3.65s

The same session in a fresh, uncached browser, without syntax highlighting: **0.42s** or about **800% faster**!

Highlighter disabled: 0.42s

I also tried to lazy-load the SyntaxHighlighter onScroll, but that makes the experience even more terrible. Instead of waiting an eternity post-load you get half a second of stuttering and erratic re-rendering in the middle of an otherwise smooth scroll - so I only went for an On/Off toggle without any "lazy" in-between option.

Yeah, it looks a bit duller, but at least we're not losing the line numbers and it does not affect the diff coloring either:
image
image image

### DIFF
--- a/src/server/db/settings.ts
+++ b/src/server/db/settings.ts
@@ -12,6 +12,7 @@ export const SETTINGS_KEYS = {
   DISPLAY_SHOW_STATS: 'display.showStats',
   DISPLAY_SHOW_AGENT_DEFINITIONS: 'display.showAgentDefinitions',
   DISPLAY_SHOW_WORKFLOW_BARS: 'display.showWorkflowBars',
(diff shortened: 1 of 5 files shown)


co-l:
I'm also investigating switching to another syntax highlighter, trying shiki right now.

PR #17, opened 2026-05-20
[Needs testing] feat: optimize session loading, fix e2e tests, some minor fixes

0x3dlux:
- Eliminate double (and triple) session loading
  - down from 2x REST + 2x Websocket at session change and 3x at page reload → 1x REST + 1x tiny ACK via Websocket
- WebSocket session.load now sends only a single ACK (instead of 2 or 3 multi-MB responses with the full session data)
- Show git status immediately on session load (eliminate 10s delay)
- Fix message count accuracy in the left sidebar (count increasing when session selected was a side-effect of loading everything 2 or 3 times)
- Add batch settings API endpoint (7 individual network calls → 1 batched call)
- Fix spinner randomly spinning after page refresh (use database is_running, not EventStore)
- Update e2e tests to fetch session via REST API after session.load
- Add context.state emission on session.load for e2e tests
- Add vitest-localstorage-mock to fix "theme"-test related issues
- Fix 4 npm vulnerabilities (2 moderate, 2 high) via npm audit fix

All 1421 unit tests and 220 e2e tests pass, but this PR has became rather crowded again because it was all somewhat related. I wouldn't be surprised if this broke some obscure corner-cases so that's why WIP + needs testing.

### DIFF
--- a/e2e/reconnection.test.ts
+++ b/e2e/reconnection.test.ts
@@ -85,17 +85,16 @@ describe('Session Reconnection', () => {
       const client2 = await createTestClient({ url: server.wsUrl })
 
       try {
-        // Load the session
-        const response = await client2.send('session.load', { sessionId })
-
-        expect(response.type).toBe('session.state')
+        // Load the session (triggers REST fetch internally)
+        await client2.send('session.load', { sessionId })
 
         const session = client2.getSession()!
         expect(session.id).toBe(sessionId)
         expect(session.projectId).toBe(projectId)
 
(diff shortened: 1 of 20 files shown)


0x3dlux:
Loading a single session - Before:
Websocket session switch - before
REST initial load - before

Switching between several sessions - After:
Websocket session switch - after

co-l:
Hey! Great optimizations. I've rebased it and merged it directly

PR #38, opened 2026-07-13
refactor: replace custom image detection with file-type library and simplify SVG detection

0x3dlux:
## Changes
- Replaced custom code detection logic with file-type library
- Simplified SVG detection to only check for `<?xml` or `<svg` start tags
- Removed `appearsToBeCode()` function that checked for programming language syntax patterns

## Rationale
- file-type library provides more reliable binary image detection
- SVG files require content-based detection since file-type may identify them as generic XML
- Simplified logic reduces complexity and potential false positives

## Testing
- All existing tests pass
- No breaking changes to API


0x3dlux:
The reason for this was that read_file sometimes, at random (as in: can not reproduce reliably) interpreted OpenFox's own .tsx files as SVG:
TSX as SNV

PR #47, opened 2026-07-14
fix: word-breaking for long text + logout redirect to home

0x3dlux:
- Added break-words to user messages, tool messages, correction messages
- Added overflow-x-auto to thinking blocks
- Added break-words to auto-prompt cards, tool call arguments/results/errors
- Added break-words to markdown paragraphs
- Fixed logout button to redirect to home page after clearing token


0x3dlux:
Before:
overlong before

After:
overlong after

PR #73, opened 2026-07-16
fix: folder selection broken on Windows (backslash paths)

Balou866:
### Symptoms
- The **Select** button did nothing (no response) when navigating **Open Project > Select Folder**
- `DirectoryBrowser` breadcrumbs rendered as a single unusable crumb.

### Root cause
The `directories` API returns native backslash paths on Windows (`C:\...\project`).
Project-name extraction in `ProjectDropdown` and `CreateSessionModal` split only on `'/'`,
so the full path (with `:` and `\`) became the project name — rejected by `validateProjectName`.
The breadcrumbs made the same separator assumption.

### Fix
- Split on both separators (`/` and `\`).
- Build breadcrumb paths using the detected separator.

### Tests
-  Path helpers extracted to web/src/lib/path.ts and unit-tested (Unix + Windows separators).

### DIFF
--- a/web/src/components/CreateSessionModal.tsx
+++ b/web/src/components/CreateSessionModal.tsx
@@ -4,7 +4,7 @@ import { useProjectStore } from '../stores/project'
 import { Modal } from './shared/Modal'
 import { Button } from './shared/Button'
 import { FolderIcon, TrashIcon } from './shared/icons'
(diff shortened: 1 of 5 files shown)


Balou866:
Added unit tests (see updated description)

PR #74, opened 2026-07-16
fix: false 'outside workdir' prompts on Windows command switches

Balou866:
### Symptoms
- Commands like `dir /s` or `findstr /i` triggered a spurious "paths outside workdir" confirmation on Windows.
- POSIX-style paths emitted by models (`/d/github/...`) were mangled the same way.

### Root cause
`extractAbsolutePathsFromCommand` treated every `/token` in a command as a Unix absolute path.
On Windows, where the shell is `cmd.exe`, command switches were extracted as paths and resolved
against the current drive (`/s` → `D:\s`). Meanwhile, genuine drive-letter paths (`C:\...`)
were never extracted at all.

### Fix
- On `win32`, extract only drive-letter paths (`C:\...` or `C:/...`).
- Unix behavior unchanged.
- As a side effect, genuine out-of-sandbox access via absolute Windows paths now prompts as intended.

### Tests
Adds platform-pinned unit tests: cmd switches ignored, drive-letter paths now extracted (closes a confirmation bypass), quoted/tilde/dedup cases.

### DIFF
--- a/src/server/tools/path-security.test.ts
+++ b/src/server/tools/path-security.test.ts
@@ -36,8 +36,16 @@ let CANONICAL_PASSWD: string
 let CANONICAL_VAR_LOG_SYSLOG: string
 let CANONICAL_ETC_ENV: string
 
+const REAL_PLATFORM = process.platform
+
+/** Pin process.platform so platform-dependent extraction logic is deterministic. */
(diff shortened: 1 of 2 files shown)


Balou866:
Added unit tests (see updated description)

PR #78, opened 2026-07-16
feat: add plugin management UI

JamesDAdams:
Hi,

I added this code to show the plugin list in the settings and give the possibility to install, uninstall, or update plugins directly from the UI.

For this, we have a `plugins-registry.json` file which stores the official list of plugins (users can still install plugins that are not in this list).

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "dist",
     "!dist/**/*.map",
     "dist/**/*.md",
-    "package.json"
+    "package.json",
+    "plugins-registry.json"
   ],
   "scripts": {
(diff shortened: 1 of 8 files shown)


arthurlacoste:
Oh that cool ! You've made another plugin ?  ✈️

The more important stuff seems to make a "one click" install, to have a sweet UI

JamesDAdams:
> Oh that cool ! You've made another plugin ? ✈️
> 
> The more important stuff seems to make a "one click" install, to have a sweet UI

Yes, I'm making a plugin for GitHub Copilot and I opened this PR because I think it's much simpler to manage plugins directly in the UI. This way, users can easily see what's installed and have a simple update button.

PR #85, opened 2026-07-17
Add per-agent model selection with ordered fallback cascades

arthurlacoste:
## Summary

Add configurable ordered model cascades for top-level agents and sub-agents.

Each agent can now either:

- inherit the active session model, which remains the default;
- define its own ordered provider/model cascade.

When a custom cascade is enabled, it fully replaces the session model for that agent. The first configured model always has priority, and fallback entries may use different providers.

## Fallback behavior

Fallback only occurs when a model fails before producing visible or actionable output.

A fallback is not attempted after receiving:

- text;
- thinking content;
- a tool-call delta.

This prevents duplicated responses and repeated tool execution.

The cascade runtime tracks temporary model unavailability in shared process memory.

Default cooldowns are:

- `429`, `503`, overload or quota failures: use `Retry-After`, otherwise 20 minutes;
- network failures, timeouts and other `5xx` errors: 2 minutes;
- `400`, `401`, `403` and `404`: unavailable until provider configuration changes.

When every configured model is unavailable, the request fails immediately with the reason and remaining delay for each model.

Cooldown durations can be changed under Settings → Advanced.

## Agent configuration

The agent editor now includes an ordered model cascade UI with:

- session-model inheritance;
- provider and model selectors;
- priority and fallback ordering;
- add, remove and reorder controls;
- explicit warnings for missing providers or models.

Cascade entries use structured `{ providerId, model }` values because model IDs may contain `/`.

## Built-in agent customization

Built-in agents such as Planner and Builder remain read-only.

Opening one shows its effective configuration and the `Duplicate & Customize` action. Duplicating creates a new custom agent with a fresh ID instead of editing or overriding the built-in ID.

This prevents Planner and Builder copies from sharing their built-in identifiers.

Existing built-in overrides can still be reset to reveal the bundled default configuration.

## Runtime changes

Agent-specific model resolution is applied to:

- top-level agent execution;
- conversation preprocessing;
- sub-agent execution;
- model capabilities and context limits;
- sampling and provider-specific settings;
- metrics and provider/model attribution.

Context limits are selected conservatively across the cascade so fallback to a smaller-context model cannot overflow.

## Error handling

LLM errors now preserve structured metadata including:

- HTTP status;
- parsed `Retry-After`;
- network, timeout and cancellation classification.

Stream cleanup also cancels the active reader, and terminal stream failures no longer emit duplicate recoverable errors.

## Testing

Added or updated coverage for:

- ordered fallback;
- cooldown expiration and priority restoration;
- `Retry-After`;
- permanent provider/model unavailability;
- no fallback after text or tool-call output;
- all-models-unavailable errors;
- successful fallback attribution;
- top-level and sub-agent cascade isolation;
- cascade persistence and validation;
- clearing cascades through explicit null metadata;
- built-in duplication with fresh IDs;
- built-in override reset;
- advanced cooldown settings;
- fallback UI messages.

Verification completed:

- `npm run typecheck` passes;
- `npm run lint` passes;
- focused agent duplication and override tests pass: 31/31;
- broad unit suite passes except for 12 existing macOS path-security assertions caused by `/tmp` resolving to `/private/tmp`.

## Follow-up work

Two improvements are intentionally left for follow-up work:

### Dynamic active model selector

When a cascade is active, update the main model selector dynamically to reflect the model currently used by the cascade runtime.

The UI should distinguish between:

- the session model;
- the configured priority model;
- the model currently used by the active cascade.

### Global all-purpose cascade

Add an optional global cascade applying to every top-level agent and sub-agent.

Expected precedence:

1. explicit agent-level cascade;
2. global all-purpose agent cascade;
3. active session model.

### DIFF
--- a/src/server/agents/model-cascade.test.ts
+++ b/src/server/agents/model-cascade.test.ts
@@ -1,0 +1,72 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import type { AgentDefinition } from './types.js'
+import type { LLMClientWithModel } from '../llm/client.js'
+
+const createRuntimeModelClient = vi.fn()
+vi.mock('../provider-manager.js', () => ({ createRuntimeModelClient }))
+
+const { resolveAgentLLMClient } = await import('./model-cascade.js')
+
+function client(model: string): LLMClientWithModel {
+  return {
+    getModel: () => model,
+    setModel: vi.fn(),
+    getProfile: vi.fn() as never,
+    getBackend: () => 'unknown',
+    setBackend: vi.fn(),
+    complete: vi.fn(),
+    async *stream() {},
+  }
+}
+
+function agent(
+  id: string,
+  subagent: boolean,
+  modelCascade?: Array<{ providerId: string; model: string }>,
+): AgentDefinition {
+  return {
+    metadata: {
+      id,
+      name: id,
+      description: '',
+      subagent,
+      allowedTools: [],
+      ...(modelCascade ? { modelCascade } : {}),
+    },
+    prompt: 'Prompt',
+  }
+}
+
+describe('resolveAgentLLMClient', () => {
+  beforeEach(() => createRuntimeModelClient.mockReset())
+
+  it('inherits the session client when no cascade is configured', () => {
+    const inherited = client('session-model')
+    expect(resolveAgentLLMClient(agent('planner', false), inherited)).toBe(inherited)
+    expect(createRuntimeModelClient).not.toHaveBeenCalled()
+  })
+
+  it('isolates top-level and sub-agent cascades', () => {
+    const plannerClient = client('planner-model')
+    const verifierClient = client('verifier-model')
+    createRuntimeModelClient.mockImplementation((providerId: string) =>
+      providerId === 'planner-provider' ? plannerClient : verifierClient,
+    )
+
+    const planner = resolveAgentLLMClient(
+      agent('planner', false, [{ providerId: 'planner-provider', model: 'planner-model' }]),
+      client('session'),
+    )
+    const verifier = resolveAgentLLMClient(
+      agent('verifier', true, [{ providerId: 'verifier-provider', model: 'verifier-model' }]),
+      client('session'),
+    )
+
+    expect(planner.getModel()).toBe('planner-model')
+    expect(verifier.getModel()).toBe('verifier-model')
+    expect(createRuntimeModelClient.mock.calls).toEqual([
+      ['planner-provider', 'planner-model'],
+      ['verifier-provider', 'verifier-model'],
+    ])
+  })
+})
--- a/src/server/agents/model-cascade.ts
+++ b/src/server/agents/model-cascade.ts
@@ -1,0 +1,15 @@
+import type { AgentDefinition } from './types.js'
+import type { LLMClientWithModel } from '../llm/client.js'
+import { createCascadingLLMClient } from '../llm/model-cascade.js'
+import { createRuntimeModelClient } from '../provider-manager.js'
+
+export function resolveAgentLLMClient(agent: AgentDefinition, inherited: LLMClientWithModel): LLMClientWithModel {
+  const refs = agent.metadata.modelCascade
+  if (!refs?.length) return inherited
+  const entries = refs.map((ref) => {
+    const client = createRuntimeModelClient(ref.providerId, ref.model)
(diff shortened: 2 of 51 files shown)


co-l:
Hey!

So, I'm not 100% convinced.

This is great maybe for cloud and OpenRouter, but the cache invalidation potential locally is huge. For API you pay money, locally you'll have a huge prefill. Not sure that's the good way to solve failure ; at least it's not a cache-efficient one.

The provider/model setter on agent/sub agent is really good though, it's a feature that's useful.

The model cascade is opt-in so not so bad at the same time, but it makes the PR do two big things and modify 56 files which in itself is scary.

Tell me what you think

arthurlacoste:
Same as with the ChatGPT PR, I think this is mostly about vision: keeping
the core lean and customizable with plugins, versus adding built-in killer
features.

The ability to choose a specific model feels like a good fit for the core.

For the cascade part, I already use reverse proxies that handle this kind
of routing between local models, remote servers, and different backends.

One use case is having a tiny model at the edge for fast tasks, with a
fallback when it fails. Apple’s local Foundation Model is a good example,
especially with its 4,000-token context. 🤡

I’m not an expert on caching, but I assume there’s no real way to avoid
cache invalidation when switching models. That’s a good point.

Could we think of fallback mainly as a safety net for cases where a small
or less reliable local model fails?

co-l:
Yeah a plugin system for that kind of behavior would be the best solution. I'm 100% open to that.

> Could we think of fallback mainly as a safety net for cases where a small
or less reliable local model fails?

I don't think that's a good idea as a general solution.
The root issue of a model fails can often be solved by automations that respect the cache. Look at "auto-retry patterns" for instance

image

That solved my main issue: being able to auto-retry on tool calls that escapes the inference engine and break the LLM loop.

A real failure is when a model is entering a loop state and the user isn't there to interrupt and take it out of it ; I'd solve that with an automation that detects 3-5 repetition of the exact same tool call or thinking or content and would abort and restart with a good guiding message "You repeated the same tool call 5 times, it's NOT WORKING BUDDY!"

arthurlacoste:
I think we can move forward like this, if you're okay with it:

* Core: allow selecting a specific model
* Plugin: handle fallback and cascade behavior

Tell me if it seems ok for you ?

The auto-retry pattern idea is really interesting, though. It is much more deterministic and probably solves a large part of the practical failure cases without changing models or invalidating the cache.

To be honest, model switching and cache invalidation lead into the same broader problem for me. This is actually the subject of my thesis.

I’m trying to find a more elegant alternative to compaction. Compaction can sometimes lose too much context. With smaller models, it can also happen too late, or the summary itself can consume most of the remaining context when the subject is complex.

The idea would be to maintain a continuous summary of the session, so it could help with several cases:

1. Avoiding or delaying compaction.
2. Preserving important context more reliably than a single late compaction step.
3. Working around a model’s context limit, for example using a model limited to 50k tokens on a task that spans 200k tokens.
4. Allowing model switching without depending on the previous model’s prompt cache.

This also matters for systems such as FreeLLMPool or FreeLLMAPI, where the backend may switch models continuously, potentially on every request. In that situation, prompt caching is already unreliable or unavailable by design. A session-level memory layer could provide continuity independently of whichever model handles the next request.

It would be somewhat similar to RAG, but applied to a conversation session:

* maintain a continuous summary;
* preserve important facts and decisions;
* retrieve older messages when needed;
* re-inject only the relevant data into the current context.

There are obviously many hard problems around deciding what to summarize, what to preserve verbatim, when to update the summary, and how to retrieve the right information without silently losing important context.

In the meantime, I’m going to do the much less glamorous but far more reasonable thing: make the compaction percentage configurable. That is considerably simpler than solving session memory and context management as a research problem inside this PR.

co-l:
>I think we can move forward like this, if you're okay with it:
> Core: allow selecting a specific model
> Plugin: handle fallback and cascade behavior
>Tell me if it seems ok for you ?

Yes that's perfect.

About compaction, unfortunately I'm cache-rich with my setup, so this makes way harder for me to think about the proper solution.

What I've seen Claude Code do is give access to the new context to the transcript of the whole context that was compacted, so the agent can grep it if required.

I'm interested in what concrete issues you encountered, that'll give me more data to think about a proper system-based solution. For now if a session is going through a workflow, the acceptance criteria is a good anchor for context, and it could probably be extended to make it even more valuable.

co-l:
I'm closing this one (clean-up)

PR #94, opened 2026-07-17
feat: improve dev mode version check and page title

0x3dlux:
## Changes

- **Page title format**: Changed from `dev- Project - Session | OpenFox` to `Project - Session | OpenFox-dev`
- **Version check**: Uses `git fetch` + `git describe` to get latest tag from `upstream/develop` instead of GitHub API
- **Caching**: 1-hour cache with `?force=true` query param to bypass
- **No polling**: Removed auto-check on mount; only runs on manual button click
- **Spinner fix**: Prevents frame-skipping by not resetting index on every render
- **Modal title**: Fixed capitalization: 'New OpenFox (dev) version available' / 'New OpenFox version available'
- **Tests**: Added 2 new tests for caching and force refresh behavior
- **Complete spinner**: 8-frame braille rotation for smooth 360° animation

### DIFF
--- a/src/server/routes/auto-update.test.ts
+++ b/src/server/routes/auto-update.test.ts
@@ -104,7 +104,7 @@ describe('Auto Update Routes', () => {
       await fetch(`${baseUrl}/api/auto-update/check`)
       mockSpawn.mockClear()
 
-      // Force refresh should call npm view again
+      // Force refresh should call git fetch again
       const res = await fetch(`${baseUrl}/api/auto-update/check?force=true`)
       expect(res.status).toBe(200)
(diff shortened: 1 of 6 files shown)


0x3dlux:
Hmm, CI still flakey there, running into timeouts. My tests seem fine after 3 attempts 🙈.

Reason for the header reformatting btw, is that I use my OS's taskbar as "project tabs" (one browser window per project) a lot, and I got kind of tired to have "dev -" eat up a 1/4 of the space there, and I really don't see much purpose in separating dev from "prod" (on a purely client-side project) beyond which update channel to query.

PR #97, opened 2026-07-18
fix(windows): make unit test suite pass on Windows

Balou866:
## Summary

Makes the full unit-test suite pass on Windows and eliminates the suite's visible side effects (orphan `cmd.exe` processes, console popups, MSYS-pipe hangs). About half the failures were product bugs the tests correctly surfaced; the rest was non-portable test code.

**Before / After (Windows 11 / Node 24, after rebase on current develop):**

```
Test Files   20 failed |  171 passed | 1 skipped   ->   193 passed | 1 skipped (194)
Tests       224 failed | 1986 passed | 3 skipped   ->  2235 passed | 11 skipped (2253)
```

## What Changed

**Product fixes**

- `src/server/utils/shell.ts` — `spawnShell()` no longer passes `detached: true` on win32; cmd.exe invocation uses `['/d', '/s', '/c', '"<command>"']` with `windowsVerbatimArguments: true`, matching Node's `{ shell: true }`. Unix keeps `detached: true`, which the process-group kill from #94 relies on.
- `src/server/tools/path-security.ts` — new `normalizeExtracted()` helper normalizes each path by its own shape (`win32.normalize` for drive-letter paths, `posix.normalize` otherwise); `isSafePath()` uses posix semantics for `SAFE_PATHS`. Strict no-op on Linux.
- `src/server/terminal/manager.ts` — tree-kill the shell **before** `pty.kill()`; `killAll()` is now async so test teardown can await it.

**Test changes**

- `src/server/tools/path-security.test.ts` — `canonicalOrResolved()` so Unix literals like `/etc/passwd` keep working as "nonexistent path outside the sandbox" instead of crashing `beforeEach`.
- `src/server/git/worktree-sync.test.ts`, `src/server/skills/registry.test.ts`, `src/server/tools/path-security.test.ts` — `CAN_SYMLINK` probe + `it.skipIf` / `describe.skipIf` (skipped without Developer Mode).
- `src/server/utils/process-tree.test.ts` — descendants via `Get-CimInstance Win32_Process` on win32; all process trees (including the new process-group tests from #94) spawned via a shared node script instead of `bash`: a PATH `bash` may be WSL, whose children live outside the host process table. Same group semantics on Unix.
- Streaming / shell tests — portable commands: `node -e "...repeat(...)"` instead of `printf` + brace expansion, `&&` instead of `;`, quoted `process.execPath`, deadline `sleep 0.1`/<100 ms → `sleep 1`/<1000 ms.

**Config**

- `vitest.config.ts` — `resolve.dedupe: ['react', 'react-dom']`, `server.deps.inline: ['zustand']`, `testTimeout: 15000` for two tests (init-llm, test-params) that flake past 5 s under full-suite load.

## Motivation & Context

I'm mainly on Windows, and every time I open a PR the test suite is not adapted to the platform — example: `npm run test:unit` on a clean `develop` checkout (71d816d) fails massively and the suite has visible side effects on the machine: orphan `cmd.exe` processes, console popups, and error dialogs such as `0x800700e8 (The pipe is being closed)` on MSYS binaries under detached spawn. This made it impossible to validate a PR by a green suite on Windows, which is the direct motivation for this PR.

## How Has This Been Tested?

- [x] Unit tests
- [ ] Integration tests
- [x] Manual testing

`npm run test:unit` after rebasing on current develop — `193 passed | 1 skipped` files, `2235 passed | 11 skipped` tests (the 11 skips are symlink tests without Developer Mode; 3 consecutive green runs pre-rebase, 1 post-rebase). `npm run typecheck:server` clean. ESLint + Prettier clean on all touched files. Manual repro: `cmd /c "echo first | tail -3"` via `spawnShell` — hung before, returns instantly now; 0 leaked `cmd.exe` / `tail.exe` after the terminal test files (previously 3–4 per run).

Unix not run locally (Windows dev machine). Every product change branches on `process.platform === 'win32'` or is a posix no-op; deferring to CI.

## Breaking Changes

None. On Linux every change is either platform-branched or a strict no-op (`posix.normalize` ≡ host `normalize`; `killAll()` await only changes timing). Test changes replace hardcoded Unix strings with `join` / `resolve` / `node -e` equivalents that resolve to the same values on Linux.

- [ ] This PR introduces breaking changes

## Related Issues

Rebased on current develop after #94 (process-group kill) landed: the branch originally carried the same `process-tree.ts` immediate-abort fix — now subsumed by #94 and dropped from this diff. The follow-up commit adapts #94's new process-group tests to a node-spawned tree so they pass on Windows too.

## AI-Enhanced Development

- **AI Models:** Claude Fable 5 (via Claude Code) — diagnosis, fixes, and PR text. Every fix was reproduced and verified on a real Windows 11 machine.

## Cache Impact

- **No** — test files, Windows-only runtime branches, and vitest config; no system prompts, tool definitions, or skills touched.

## Checklist

- [x] My code follows the project style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code where necessary
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [ ] I have updated documentation if needed

**Key files to review:** `src/server/utils/shell.ts`, `src/server/tools/path-security.ts`, `src/server/terminal/manager.ts`, `vitest.config.ts`

---

<details>
<summary><b>Technical investigation</b> — symptoms, root cause, and fix details</summary>

### Symptoms

Clusters of failures and their errors, verbatim:

- `path-security.test.ts` — 174 failures, all in `beforeEach`:
  ```
  ENOENT: no such file or directory, lstat 'D:\etc\passwd'
  ```
- `process-tree.test.ts`:
  ```
  Error: Command failed: ps -eo pid=,ppid=
  ```
- `worktree-sync.test.ts`, `skills/registry.test.ts`:
  ```
  Error: EPERM: operation not permitted, symlink '..\some-package\index.js' -> '...\node_modules\.bin\some-package'
  ```
- All web component tests (PluginsTab, AdvancedTab, CriteriaEditor, DirectoryBrowser, SkillsModal, ...):
  ```
  TypeError: Cannot read properties of null (reading 'useState')
  ```
- Assorted path-shape failures, e.g.:
  ```
  AssertionError: expected 'D:\tmp\project\worktrees\feature-test' to contain 'worktrees/feature-test'
  ```

### Cause

About half the failures were product bugs the tests correctly exposed; the other half was non-portable test code.

**Product bugs**

1. **`spawnShell()`** (`src/server/utils/shell.ts`) passed `detached: true` through to `spawn()` on win32. A detached child gets its own console (hence the popup windows), and MSYS binaries (`tail`, `sleep` from Git for Windows) hang or fail on their pipes in that configuration — empirically, `cmd /c "echo first | tail -3"` never returns with `detached: true` and completes instantly without it. Nothing needs `detached` on win32: tree-kill goes through `taskkill`, not process groups.
2. **Same function** — Node's default argument quoting escapes inner double quotes with backslashes, which cmd.exe does not understand. Any command containing quotes (e.g. `node -e "..."`) broke silently:
   ```
   '\"C:\Program Files\nodejs\node.exe\"' n'est pas reconnu en tant que commande interne...
   ```
3. **`path-security.ts`** normalized extracted command paths with the **host** `normalize()`, turning `/var/log` into `\var\log` on Windows. Side effect: `isSafePath()` could never match `SAFE_PATHS` entries like `/dev/null` on win32.
4. **`terminal/manager.ts`** — under ConPTY, `pty.kill()` leaves the spawned `cmd.exe` alive, and `killAll()` was fire-and-forget so test workers exited before cleanup finished — orphan cmd.exe accumulated with every run. Killing the shell first also reparents its children out of `taskkill /t`'s reach, so the tree must be killed *before* `pty.kill()`.

**Test-side causes:** hardcoded Unix paths (`/etc/passwd`, `/tmp/project/...`), `realpath()` on paths that do not exist on Windows, unprivileged symlink creation (needs Developer Mode on win32), bash-only syntax (`printf 'X%.0s' {1..N}`, `;` separators), unquoted `process.execPath` (contains a space on Windows), process trees spawned via `bash` (a PATH `bash` may be WSL, whose children are invisible to the host process table), and duplicate React copies between the root and `web/node_modules` trees breaking hooks in every web test (`zustand` is externalized by vitest, so its own `import 'react'` bypassed dedupe — hence `server.deps.inline`).

</details>

### DIFF
--- a/src/server/git/worktree-config.test.ts
+++ b/src/server/git/worktree-config.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { join } from 'node:path'
 import { loadWorktreeConfig, saveWorktreeConfig } from './worktree-config.js'
 
 vi.mock('node:fs/promises', () => ({
@@ -58,7 +59,7 @@ describe('saveWorktreeConfig', () => {
 
     expect(mkdir).toHaveBeenCalledWith(expect.stringContaining('.openfox'), { recursive: true })
     expect(writeFile).toHaveBeenCalledWith(
-      expect.stringContaining('.openfox/worktree.json'),
+      expect.stringContaining(join('.openfox', 'worktree.json')),
       JSON.stringify({ ignoredAssets: 'copy', overrides: { node_modules: 'symlink' } }, null, 2) + '\n',
       'utf-8',
     )
--- a/src/server/git/worktree-sync.test.ts
+++ b/src/server/git/worktree-sync.test.ts
@@ -8,6 +8,19 @@ vi.mock('../utils/logger.js', () => ({
   logger: { debug: vi.fn(), warn: vi.fn(), info: vi.fn(), error: vi.fn() },
 }))
 
+/** Symlink creation needs a privilege/Developer Mode on Windows — probe once. */
+const CAN_SYMLINK = (() => {
+  if (process.platform !== 'win32') return true
+  const probe = join(tmpdir(), `wt-symlink-probe-${process.pid}`)
+  try {
+    symlinkSync('probe-target', probe, 'file')
+    rmSync(probe)
+    return true
+  } catch {
+    return false
+  }
+})()
+
 function createProjectStructure(root: string) {
   // Create a realistic project with .gitignore, node_modules, etc.
   mkdirSync(join(root, '.git'))
@@ -17,13 +30,16 @@ function createProjectStructure(root: string) {
   mkdirSync(join(root, 'node_modules', '.bin'), { recursive: true })
   mkdirSync(join(root, 'node_modules', 'some-package'), { recursive: true })
   writeFileSync(join(root, 'node_modules', 'some-package', 'index.js'), 'module.exports = 42')
-  symlinkSync('../some-package/index.js', join(root, 'node_modules', '.bin', 'some-package'))
 
   // Create web/node_modules
   mkdirSync(join(root, 'web', 'node_modules', '.bin'), { recursive: true })
   mkdirSync(join(root, 'web', 'node_modules', 'web-pkg'), { recursive: true })
   writeFileSync(join(root, 'web', 'node_modules', 'web-pkg', 'index.js'), 'module.exports = "web"')
-  symlinkSync('../web-pkg/index.js', join(root, 'web', 'node_modules', '.bin', 'web-pkg'))
+
+  if (CAN_SYMLINK) {
+    symlinkSync('../some-package/index.js', join(root, 'node_modules', '.bin', 'some-package'))
+    symlinkSync('../web-pkg/index.js', join(root, 'web', 'node_modules', '.bin', 'web-pkg'))
+  }
 }
 
 describe('getIgnoredDirectories', () => {
@@ -46,7 +62,7 @@ describe('getIgnoredDirectories', () => {
 })
 
 describe('syncIgnoredAssets', () => {
-  it('copies node_modules preserving symlinks', async () => {
+  it.skipIf(!CAN_SYMLINK)('copies node_modules preserving symlinks', async () => {
     const root = mkdtempSync(join(tmpdir(), 'wt-test-'))
     const worktreePath = join(root, 'worktrees', 'test-branch')
     try {
--- a/src/server/git/worktree.test.ts
+++ b/src/server/git/worktree.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { join, resolve } from 'node:path'
 import {
   getGitBranch,
   listWorktrees,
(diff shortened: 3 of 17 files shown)


co-l:
That's awesome!

I'm fixing the CI tests, but it's a WIP. I also added a ai-disclosure requirement, to know what model was used for the PR.

Balou866:
Thanks! 
Rebased on develop to clear the conflicts with #94, hence the force-push.

Two side effects: I dropped my process-tree.ts fix — #94 already covers it. 
And I adapted your new process-group tests in a small follow-up commit: they spawn their tree with bash, which on Windows can be WSL bash, so the children are invisible to the host and the assertions fail. Same tests, just node-spawned now.

FYI, my spawnShell change only drops detached on win32 — Unix keeps it, so your group-kill path is untouched.
Suite still green on Windows after the rebase (2235 passed). 

Description updated to the new template. 
I had to use Fable 5 on this one (I have some credits to spare till tomorrow), this was a way more bigger task than what I had anticipated 😂

co-l:
I can only imagine haha, going from 20 failing test files to 100% success, that's no small task.

Ok you're now the official Windows reference contributor. I'll merge that asap.

Issue #108, opened 2026-07-18
Some edit file tool call make the workflow stop


pillisan42:
What you have done with this project is pretty amazing I like it a lot 

I already told that I had a problem with an agent getting stuck in JadeDAdams but I think my problem is different

In openfox it displays running but in lmstudio the model does nothing and it stop right after a fail tool call

<img width="450" height="446" alt="Image" src="https://github.com/user-attachments/assets/a2eb7848-5306-42aa-8fd4-9ffc973daec1" />

This is not always but it’s pretty regular and it seems I need to restart a complete session If i want to go through it

I analyze it more thoroughly since the problem is systematic here is the log from lm studio with tool calling
``` 
timestamp: 18/07/2026 19:15:21
type: llm.prediction.output
modelIdentifier: qwen3.6-27b-mtp
output:
Now I need to modify the heartbeat functions to return VersionCheck. The key changes:
1. `heartbeat()` returns `Result<VersionCheck>` instead of `Result<()>`
2. `heartbeat_with_retry()` returns `Result<VersionCheck>` instead of `Result<()>`
3. `heartbeat_with_cancel()` returns `Result<VersionCheck>` instead of `Result<()>`

Then in main.rs, I need to:
1. Use the VersionCheck from heartbeat to trigger auto-update
2. Add platform detection and version comparison
3. Wire up the auto-update flow

Let me also add the telemetry endpoint for update results (C20).

Let me do this step by step. First, C13 - network/mod.rs changes.
</think>

Now implementing C13 - modifying heartbeat functions to return VersionCheck.

<tool_call>
<function=edit_file>
<parameter=path>
pudelo-agent/src/network/mod.rs
</parameter>
<parameter=old_string>
    /// Heartbeat: inform the server this machine is alive (single attempt, no retry).
    pub fn heartbeat(
        &self,
        machine_id: &str,
        hostname: &str,
        os_type: &str,
        os_version: &str,
    ) -> Result<()> {
        let agent_version = env!("CARGO_PKG_VERSION");
        debug!("POST {}/api/v1/agent/heartbeat (machine={}, host={})",
            self.config.url, machine_id, hostname);
        let response = self.client.post(format!(
            "{}/api/v1/agent/heartbeat",
            self.config.url
        ))
            .header("X-Machine-ID", machine_id)
            .header("X-Hostname", hostname)
            .header("X-OS-Type", os_type)
            .header("X-OS-Version", os_version)
            .header("X-Agent-Version", agent_version)
[...]

PR #117, opened 2026-07-19
fix: send empty string instead of null for assistant tool-call messages

RenZan:
## Description

After a failed `edit_file` tool call, the agent loop would stop responding when using LM Studio / Qwen backends. The model shows no output despite OpenFox displaying "running".

## Root Cause

`buildAssistantMessage()` in `src/server/llm/client-pure.ts` serialized assistant messages with `tool_calls` using `content: null`. LM Studio / Qwen chat templates mishandle `null` content for tool-call turns, causing the model to silently stop generating after receiving a failed tool result.

## Fix

Changed `buildAssistantMessage` to send `content: ` (empty string) when the assistant message has `tool_calls`, while keeping `content: null` for plain text responses.

## Changes

- `src/server/llm/client-pure.ts` — core fix
- `src/server/llm/client-pure.test.ts` — updated assertions
- `src/server/chat/agent-loop.integration.test.ts` — integration test for failed tool -> recovery flow

## Validation

- `typecheck`: clean
- `lint`: clean  
- `test:unit`: 2246 passed, 3 skipped

cc @pillisan42 — tentative fix, could you test if this resolves the issue?

## AI Model Disclosure

DeepSeek V4 Flash

Try closing #108

### DIFF
--- a/src/server/chat/agent-loop.integration.test.ts
+++ b/src/server/chat/agent-loop.integration.test.ts
@@ -365,4 +365,58 @@ describe('agentLoop integration', () => {
       .filter((e: any) => e.type === 'chat.done' && e.data?.reason === 'complete')
     expect(chatDoneEvents.length).toBeGreaterThanOrEqual(1)
   })
+
+  it('continues loop after a failed tool call and allows recovery in next iteration', async () => {
+    const append = vi.fn()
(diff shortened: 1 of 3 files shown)


pillisan42:
@RenZan Thanks for that i will try it when merge and give you a feedback

PR #121, opened 2026-07-19
fix(migration): crash on upgrade from 2.0.70 to 2.0.75. Update columnNames after renaming worktree to workspace in migration.

bruspal:
### Summary

After renaming column name from `worktree` to `workspace` columnNames is not synchronized with the real table structure. Overiding `worktree` with `workspace` should fix the following `if (!columnNames.includes('workspace'))` to prevent to recreate a column which has just been renamed.

### AI-Enhanced Development

- **AI Models:** DeepSeek V4 Flash (API)

### Cache Impact

- **No**


bruspal:
Un petit fix sur la migration lors du passage de la 2.0.70 à la 2.0.75

PR #134, opened 2026-07-21
feat(context): add configurable auto-compaction

arthurlacoste:
### Summary

This PR adds configurable automatic context compaction and makes its behavior visible in both the session menu and Advanced Settings.

Auto-compaction is now configured as a ratio internally and displayed as a percentage in the UI:

```text
0%   = disabled
85%  = compact at 85% of the model context window
100% = 1.0
```

The value can be configured through:

```text
Settings > Advanced
Session menu > Compact
OPENFOX_COMPACTION_THRESHOLD
```

The environment variable keeps ratio semantics:

```bash
OPENFOX_COMPACTION_THRESHOLD=0.85
```

When the environment variable is set, it remains authoritative and the UI controls are disabled.

#### Compact menu

The compact menu now exposes the threshold directly below the manual compact action:

```text
┌──────────────────────────────────────┐
│ Compact                              │
├──────────────────────────────────────┤
│ Auto-compact              85% · 170k │
│ [────────────────────●─────────────] │
│ Minimum: 14% · ~28k tokens           │
│                                      │
│ 14% Full · estimated breakdown       │
│ [System][Tools][Skills][MCP][Free  ] │
│                                      │
│ System prompt          ~10k · 5.0%   │
│ Built-in tools          ~8k · 4.0%   │
│ MCP tools               ~6k · 3.0%   │
│ Skills                  ~2k · 1.0%   │
│ Current conversation    ~2k · 1.0%   │
│ Free context           ~172k · 86.0% │
└──────────────────────────────────────┘
```

The menu is wider than before so the slider and token budget remain readable.

#### Advanced Settings

Advanced Settings contains a larger proportional view of the complete model context:

```text
Auto-Compaction                         85% · 170k

[────────────────────────●──────────────]

Minimum: 14% · ~28k tokens

0%                                            100%
┌───────────────────────────────────────────────┐
│ System │ Tools │ Skills │ MCP │ Conversation │ Free
└───────────────────────────────────────────────┘
```

The graph distinguishes:

- system prompt
- custom instructions
- skills
- subagent definitions
- built-in tools
- MCP tools
- current conversation
- free context

Each segment shows an estimated token count and its percentage of the selected model context. Small segments remain identifiable through hover details.

The minimum slider value corresponds visually to the sum of the incompressible segments. The slider accepts either `0%` to disable automatic compaction or a value at or above the session-specific minimum.

#### Dynamic model context

The graph and threshold refresh when:

- a model is selected
- the active model changes
- a session-specific model changes
- the configured model context window changes
- dynamic context is applied
- instructions, skills, tools, or MCP tools change
- context usage changes during a conversation
- compaction completes

No message needs to be sent before the graph appears. If no model is selected, the graph stays hidden because there is no reliable context-window size to use.

#### Compaction safety

The compaction loop now prevents recursive summary generation.

Only one automatic compaction can occur during a single agent run, even if the resulting summary still exceeds the configured threshold. This prevents the agent from repeatedly summarizing its own compaction summary and consuming more tokens than it saves.

After compaction, context usage is immediately re-estimated from:

```text
incompressible context + generated summary
```

so the UI no longer displays the pre-compaction usage until the next model request.

#### Context floor calculation

The incompressible floor is calculated from the same cached prompt and tool definitions used for the actual LLM request.

Prompt construction now exposes its real constituent parts instead of reconstructing categories afterward:

```text
system
instructions
skills
subagent definitions
```

Tool definitions are separated into:

```text
built-in tools
MCP tools
```

The calculated floor is stored alongside the cached prompt, preventing the graph from mixing cached prompt data with newer unapplied dynamic context.

Token values are explicitly presented as estimates because they currently use a model-independent character-based estimate rather than the selected model tokenizer.

#### Configuration cleanup

The unused `compactionTarget` configuration and `getCompactionTarget()` helper were removed.

Configuration persistence now includes:

```ts
context: {
  compactionThreshold: 0.85
}
```

The server also validates that the configured ratio remains between `0` and `1`.

#### Validation

```text
124 tests passed
1 test skipped

Server typecheck: passed
Web typecheck: passed
Lint: passed
Server build: passed
Web build: passed
git diff --check: passed
```

### AI-Enhanced Development

- **AI Models:** GPT-5.6 Thinking (et un peu de cervelle humaine)

### Cache Impact

- **Yes** — The cached prompt now stores its estimated compaction floor and segment breakdown alongside the system prompt, tool definitions, and dynamic-context hash. Prompt construction was also split into explicit sections so the displayed token budget matches the exact cached prompt used by the agent.


co-l:
Hello Arthur!

Good intention, configurability is great.

But it's doing too much and it's too complex:

1. Compaction floor -> a sensitive default would be 15K tokens, no need to have a complex approach for a value that's not very dangerous. If the user sets a value that's very low, the only thing they'll encounter is the direct consequence of their action, auto-compact on loop. ; same for the "once per run" guard. I don't believe those are realistic scenario, and if they are they're harmless
2. No ceiling, max value 95%, but that's where the real danger is: providing a window for outputTokens that is too small for a real summary will be costly
3. Updating context size in the UI post compaction -> currently we just wait for the first call after compact ; PR adds complexity for a transient estimate
4. No "default" button, changing the value keeps you guessing what was there before
5. Set as a global setting, while it's rather a per model setting for fine-tuning compaction behavior. 

"After review, green clanker covered his mouth, official decision is red card" (sorry that's a world cup joke)

More seriously, after analysis, only good thing I'd keep as-is is the clean up of dead code :( sorry

arthurlacoste:
Thanks for the detailed review. I went back through the whole implementation and reworked it to follow the points you raised, rather than only patching the visible parts.

The branch has now been simplified around the following decisions:

- the threshold is now stored per model instead of globally
- the default is 85%
- the maximum is capped at 95% to preserve enough room for summary output tokens
- the minimum is a simple fixed 15k-token floor
- a Default button removes the model override and restores the default value
- the post-compaction context estimate was removed
- the once-per-run guard was removed
- the dynamic compaction-floor calculation was removed
- the detailed floor breakdown and colored category graph were removed
- the dead `getCompactionTarget()` helper was removed

So the implementation no longer tries to estimate or display system prompt, tools, skills, MCP tools, subagents, and other context categories separately.

The remaining visual question is only about presentation, without reintroducing that complexity:

```text
Option A — minimal
[ used context ][ free context ]

Option B — richer visual bar
A colored proportional bar similar to the previous UI, but only using already-known values such as used context, free context, and the selected threshold. No detailed prompt-floor calculation or category estimation.
```

Which presentation would you prefer, or would you rather keep only the slider and token values with no additional bar?

The updated branch is based on `develop` and has been force-pushed with the rewritten implementation.

Edit : j'ai fait une autre PR je crois que je ne peux pas push sur une PR fermé

PR #157, opened 2026-07-22
fix(providers): preserve advanced model params when reopening provider modal

feanor91:
### Summary
- Reopening the "Manage providers" edit modal for a configured model dropped `temperature`, `topP`, `topK`, `maxTokens`, and `compactionThreshold` from local state, since the reset-on-open effect never copied them from the loaded model config.
- This made the sliders/inputs silently reset to defaults on reopen, and saving again from that state would overwrite the previously persisted values with `undefined` — i.e. changing the auto-compaction threshold (or temperature/topP/topK/maxTokens) appeared not to save.

### AI-Enhanced Development

- **AI Models:** Claude Sonnet 5 (Claude Code)

### Cache Impact

- **No**

### Test plan
- [x] Added a regression test (`preserves previously-saved advanced parameters when reopening the modal`) that renders the modal with a pre-configured model and asserts the save payload preserves temperature/topP/topK/maxTokens/compactionThreshold when reopened without touching any field.
- [x] Verified the new test fails without the fix (reverted the fix locally, confirmed `undefined` instead of the expected values) and passes with it.
- [x] Full `ProviderModal.test.tsx` suite passes (10/10).

### DIFF
--- a/web/src/components/shared/ProviderModal.test.tsx
+++ b/web/src/components/shared/ProviderModal.test.tsx
@@ -210,6 +210,56 @@ describe('ProviderModal - thinkingLevel persistence', () => {
     }
   })
 
+  it('preserves previously-saved advanced parameters when reopening the modal', async () => {
+    const editProvider = {
+      id: 'test-provider',
+      name: 'Test Provider',
+      url: 'http://localhost:8000/v1',
+      backend: 'vllm' as const,
+      models: [
+        {
+          id: 'test-model',
+          contextWindow: 200000,
+          thinkingEnabled: true,
+          temperature: 0.42,
+          topP: 0.9,
+          topK: 40,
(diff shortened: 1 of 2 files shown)


feanor91:
Hi,
During this modification I noticed a "kind of" bug, Temperature, top K, Top P cames from a hard coded table depending of the model name. and are only placeholders, i fI entered values into those fields, I have to start from 0 wirth arrows on the right.
Not sure how to treat that. My first idea was to get default values directly from the provider, but, my second thoughts was : "if you dit like that, it must be because you have a good reason"
For me it is always better to have real values in fields that can be filled.
In me same time, I don't know all providers, so, I'm not sure of potential impact if we modify values, present or not, so I preffered to not touched this .

What do you think?

PR #163, opened 2026-07-23
Feat/edit proxy

JamesDAdams:
### Summary

Make the HTTP Proxy setting (`network.proxyUrl`) apply to all OpenFox network requests, not just LLM API calls. Previously, `proxyFetch()` was used selectively in a few modules. Now `globalThis.fetch` is overridden at startup to route through the configured ProxyAgent, so every `fetch()` call in the server benefits automatically.

Additionally:

- Terminal/child processes now receive `HTTP_PROXY`/`HTTPS_PROXY` environment variables

- A "Test" button and `/api/proxy/test` endpoint allow verifying proxy connectivity from Settings → Advanced

- The old `proxyFetch()` function is removed since the global override replaces it

### AI-Enhanced Development

- **AI Models:** DeepSeek V4 Flash

### Cache Impact

- **Yes** — No cached data changed, but the proxy is no longer cached per-request: `getProxyAgent()` now calls `getSetting()` on every `fetch()` to support runtime changes. The ProxyAgent instance itself is cached and invalidated when the URL changes.

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -11,9 +11,10 @@ import type { ServerHandle } from './context.js'
 import type { VisionBackend } from './llm/vision-fallback.js'
 import { initDatabase } from './db/index.js'
 import { initEventStore } from './events/index.js'
+import './llm/proxy.js'
 import { detectModel, getLlmStatus, getBackendDisplayName } from './llm/index.js'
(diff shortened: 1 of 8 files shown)


co-l:
Hello, beaucoup de conflits, je ne sais pas si c'est normal. Peux-tu rebaser sur develop ?

PR #168, opened 2026-07-24
feat: per-session and per-project MCP server toggles

JamesDAdams:
### Summary

Add per-session and per-project MCP server toggle system with 3-layer resolution (Global → Project → Session).

**Key changes:**

- McpManager always connects servers regardless of `disabled` flag (disabled becomes a visibility filter at tool-definition level, not a connection blocker)

- Session MCP overrides: new `PUT /api/sessions/:id/mcp/overrides` endpoint, stored in DB via `mcp_disabled_servers` column, survives restarts

- Project MCP overrides: `mcpOverrides` field in WorkspaceConfig (`.openfox/workspace.json`), with expandable server card UI and individual tool toggles in ProjectSettingsModal

- Chat McpSelector dropdown now uses session-level API instead of global API — toggling only affects the current conversation

- Tool definitions filtered per-session via `getToolRegistryForAgent(sessionId)`: MCP tools from session-disabled servers are excluded from LLM tool definitions

- New sessions inherit project mcpOverrides (or global `config.disabled` as fallback)

### AI-Enhanced Development

- **AI Models:** Claude 3.5 Sonnet

### Cache Impact

- **Yes** — Tool definitions sent to the LLM now depend on session MCP overrides, affecting the cached tool fingerprint for sessions with overridden MCP server states.

### DIFF
--- a/src/cli/config.ts
+++ b/src/cli/config.ts
@@ -124,6 +124,7 @@ const mcpServerSchema = z.object({
   disabledTools: z.array(z.string()).optional(),
   cachedTools: z.array(cachedToolSchema).optional(),
   timeout: z.number().positive().optional(),
+  disabled: z.boolean().optional(),
 })
 
 const llmConfigSchema = z.object({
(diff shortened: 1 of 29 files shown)


co-l:
Hey!

I made sure the cache invalidation logic is always validated by the user, just the like the original MCP global config.

The per-session MCP is interesting, but I personally don't need it ; I kept it but it's now an opt-in option in the tools tab.

PR #173, opened 2026-07-25
fix(security): use execFileSync instead of execSync in deleteWorkspace

bruspal:
### Summary

execSync interpolates user input directly into shell commands, making it vulnerable to shell injection attacks. For example, a workspace named 'test; touch ~/.pwned' would execute 'touch ~/.pwned' via the shell.

execFileSync passes arguments directly to the process without shell interpretation, preventing any command injection regardless of workspace name content. The fallback to fs.rm() in the catch block ensures robust deletion even if rm fails for unexpected reasons.

The dynamic import of 'rm' was moved into the catch block because it's only needed as a fallback. This keeps the happy path (execFileSync succeeds) free from unnecessary module loading overhead.

### AI-Enhanced Development

- **AI Models:** qwen3.6-27b, ornith-1.0-35b

### Cache Impact

- **No**

### DIFF
--- a/src/server/git/workspace.test.ts
+++ b/src/server/git/workspace.test.ts
@@ -11,17 +11,20 @@ import {
   listWorkspaces,
   getWorkspacesDir,
   isGitRepository,
+  deleteWorkspace,
 } from './workspace.js'
 
 vi.mock('node:child_process', () => ({
   spawn: vi.fn(),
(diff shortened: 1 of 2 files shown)


bruspal:
Hello,

mauvaise branche pour cible. Je refais sur la branche de dev.

PR #178, opened 2026-07-28
Fix systemd service target

bruspal:
### Summary

The actual service does not starts on headless system because of its dependancy with graphical interface.
Change `graphical-session.target` to `default.target`.
`graphical-session.target` never hit on headless / console only systems.

While openfox as service is a pure cli application I removed the X11/XWayland pulling in generated start.sh script.

I split changes in two commits :
- **first** : only change target. **Compulsary for starting on headless.**
- **second** : **Optional.** Cleanup of start.sh, remove useless code with new service settings.
  Should have no impacts unless I missed some details.

---

- Unit test pass, no need to change test.
- Manual test tests are OK on headless system, no impact on system with X11/Wayland PC.

**Full details in commit messages.**

### AI-Enhanced Development

Helps to quickly find service management code and makes implémentations faster.

- **AI Models:** Qwen-35B-A3B

### Cache Impact

- **No**

### DIFF
--- a/src/cli/service.ts
+++ b/src/cli/service.ts
@@ -115,8 +115,7 @@ async function createSystemdService(): Promise<void> {
   const servicePath = expandPath(SERVICE_PATH)
   const serviceContent = `[Unit]
 Description=OpenFox Agentic Coding Assistant
-After=graphical-session.target
-Wants=graphical-session.target
+After=default.target
 
 [Service]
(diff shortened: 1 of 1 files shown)


bruspal:
oops j'ai encore oublié, j'ai commit dans main...

Désolé. Si besoin je refais.

PR #191, opened 2026-08-01
fix: enforce workflow git context integrity

### LINKED ISSUE #183: Potential regression: Dev & Verify writes on main before switching to the session branch
## Environment

- OpenFox: `2.0.102`
- OS: Debian Linux
- Workflow: Plan → Add Criteria → Dev & Verify → Commit & Push
- Setting `Confirm on workspace & git actions`: disabled

## Problem

A feature branch was created during planning, but after Dev & Verify completed,
Git still reported that the repository was on `main`:

```text
## main...origin/main
 M src/data/models.py
 M src/services/pdp_simplified_draft_generator.py
?? src/data/audit.py
```

The requested branch existed but still pointed to the base commit:

```text
pilote/495-centraliser-codes-audit 2c2cf55
```

OpenFox nevertheless completed all acceptance criteria without reporting a
workspace or branch mismatch.

During Commit & Push, OpenFox finally attached the work to the correct branch:

```text
b6a038c
HEAD -> pilote/495-centraliser-codes-audit
origin/pilote/495-centraliser-codes-audit
```

No commit or push was made to `main`.

## Expected behavior

Before the first write in Dev & Verify, OpenFox should ensure that:

- the session has an effective workspace;
- the expected branch is checked out;
- the active Git branch matches the session branch.

If not, Dev & Verify should switch branches, ask for confirmation, or stop with
a clear error.

## Impact

The final push was correct, but the intermediate state is unsafe:

- an interruption leaves modifications while Git reports `main`;
- a user or external tool could commit them manually on `main`;
- verification can succeed without validating the actual Git context.

## Relation to existing issue

This is related to #150 and PR #118, but differs from a stale UI label:

- the UI showed `main`;
- Git also reported `main`;
- the feature branch existed but was not active during Dev & Verify.

## Reproducibility

Observed once on OpenFox `2.0.102`.

The confirmation setting was disabled, which explains the absence of a prompt,
but branch consistency should still be enforced independently.

### LINKED ISSUE #190: Refresh agent context after workspace or branch mutation during a turn
## Parent / related

- Parent: #183

## Problem

Git, the session and the frontend can be correctly refreshed while the agent is still reasoning on the old workspace or the old branch. After an official mutation (workspace switch, branch checkout, etc.) the agent that is mid-turn keeps seeing the stale workspace path and stale branch label in tool results, system prompt fragments and downstream reasoning, so it can keep writing files against the wrong tree.

## Observed reproduction

Seen on OpenFox `2.0.104`. A workspace or branch change is performed mid-turn (e.g. via the `workspace` / `git` tools or a workflow-level mutation). Git, the persisted session, and the frontend are updated to the new state, but the same-turn tool responses and the next-turn context the agent receives still describe the previous workspace and branch. Verification in that turn can therefore succeed against the wrong tree without any error.

## Expected behavior

- After an official workspace or branch mutation, the agent must receive an authoritative, post-mutation tool result describing the current workspace and current branch.
- The context used in the same turn (continuation prompt, tool observations, follow-up model call) must reflect that authoritative state.
- The context used at the start of the next turn must also reflect the current workspace and current branch.
- If the mutation itself fails, the agent must not be presented with a refreshed, fake post-mutation context.

## Out of scope

- Automatic checkout / branch switching on agent demand (already covered by #183).
- Frontend-only display of the refreshed state when the backend has not actually mutated.

## Acceptance criteria (tests)

1. **Same-turn refresh** — after an official workspace or branch mutation within a turn, the agent's same-turn tool observation, continuation prompt, and follow-up model call all describe the post-mutation workspace and branch.
2. **Next-turn refresh** — at the start of the next turn, the agent context (system prompt fragments, available tools' workspace metadata, etc.) reflects the current workspace and branch, not the pre-mutation values.
3. **Mutation failure** — when the official mutation fails (e.g. checkout of a non-existent branch), the agent must not be served a refreshed, fake post-mutation context; it must see the failure and the pre-mutation state.
4. **No false refreshed context** — there is no codepath that reports a refreshed context when no authoritative mutation actually happened.

theshwal:
## Summary

Enforce git-context integrity for write workflows (#183) and refresh agent
state after any official workspace/branch mutation during a turn (#190).

When the session is bound to a Git branch (e.g. through Dev & Verify or the
workspace tool) but the effective workdir cannot be verified on that
branch, the workflow is now blocked **before** any agent step, any shell
step, or any file write. Non-git projects and sessions without an explicit
branch expectation are not blocked.

After an authoritative workspace or branch mutation, the cached system
prompt is invalidated so the next LLM call (same-turn continuation AND next
user turn) rebuilds against the new workspace/branch. On mutation failure
the agent sees the failure and the pre-mutation state — there is no fake
refreshed context.

### Workflow gate (#183) — `executeWorkflow()`

- New `SessionManager.assertExecutionGitContext()` reads the effective
  workdir, the persisted session branch and the actual Git branch, and
  returns either `{ ok: true }` or `{ ok: false, reason }`.
- `executeWorkflow()` now calls the gate as the very first step (before
  start-condition evaluation, before the workflow-started marker, before
  any agent or shell step). On failure the workflow is **BLOCKED** with a
  clear reason, no agent is invoked, no checkout is performed, and
  `startWorkflow()` is never called.
- The reason explicitly includes `workdir`, the expected branch, the actual
  branch and a remediation hint pointing the user at the workspace tool.

### Fail-closed decision matrix

| `expectedBranch` | `actualBranch`           | Result     |
| ---------------- | ------------------------ | ---------- |
| `null`           | `null`                   | `{ ok: true }` |
| `null`           | `<any non-null>`         | `{ ok: true }` |
| `<set>`          | equal `<set>`            | `{ ok: true }` |
| `<set>`          | different `<set>`        | `{ ok: false, reason }` |
| `<set>`          | `null` (detached HEAD, …) | `{ ok: false, reason }` |

The last row was previously fail-open (returning `{ ok: true }` whenever
the actual branch could not be resolved) and is now `BLOCKED`. The reason
explicitly states that the actual branch is unavailable and that no agent,
no checkout and no file were performed.

### Post-mutation context refresh (#190) — `SessionManager.switchWorkspace()`

- New `SessionManager.clearCachedPrompt()` (and the matching
  `clearSessionCachedPrompt` DB helper) invalidates the cached system
  prompt, resets the warmup flag and sets `dynamicContextChanged = true`.
- `SessionManager.switchWorkspace()` calls `SessionManager.clearCachedPrompt()`
  **after** the new workdir/branch has been read authoritatively and
  **only on a successful mutation** — no fake refreshed context on failure.
- The workspace tool's success path continues to surface the authoritative
  post-mutation `workspace`, `path` and `branch`; on failure it propagates
  the underlying error without inventing state.

### Non-goals

- No frontend changes were required for this delivery. The web client
  already re-renders from the same authoritative events the server emits.

## Validation

- `npx vitest run src/server/workflows/executor.execution-context.test.ts src/server/session/manager.execution-context.test.ts src/server/workflows/executor.test.ts src/server/workflows/executor-mode.test.ts src/server/workflows/executor-resume-nudge.test.ts src/server/session/manager.test.ts src/server/tools/workspace.test.ts`
  → **Test Files 7 passed (7), Tests 136 passed (136)** — exit 0
- `npx tsc --noEmit` → clean — exit 0
- `npm run lint` → clean — exit 0
- `npx vitest run src/server src/shared`
  → **Test Files 151 passed | 2 skipped (153), Tests 2118 passed | 26 skipped (2151)** — exit 0

## Commits

- `54d70b43` — `fix(workflows): guard execution git context`
- `e1aa1e3b` — `fix(context): refresh agent state after workspace mutation`
- `8b7442c8` — `fix(workflows): fail closed when expected branch is unavailable`

## Related issues

Issue #190 was identified while investigating #183 and covers agent-context
refresh after workspace or branch mutation.

Fixes #183
Fixes #190

### DIFF
--- a/src/server/db/sessions.ts
+++ b/src/server/db/sessions.ts
@@ -14,8 +14,7 @@ function getProjectDangerLevel(projectId: string): DangerLevel {
   try {
     const db = getDatabase()
     const row = db.prepare('SELECT danger_level FROM projects WHERE id = ?').get(projectId) as
-      | { danger_level: string | null }
-      | undefined
+      { danger_level: string | null } | undefined
     return (row?.danger_level as DangerLevel) ?? 'normal'
   } catch {
     return 'normal'
@@ -235,6 +234,17 @@ export function updateSessionCachedPrompt(
   ).run(systemPrompt, JSON.stringify(tools), hash, now, id)
 }
 
+export function clearSessionCachedPrompt(id: string): void {
+  const db = getDatabase()
+  const now = new Date().toISOString()
+
+  db.prepare(
+    `
+    UPDATE sessions SET cached_system_prompt = NULL, cached_tools = NULL, cached_hash = NULL, updated_at = ? WHERE id = ?
+  `,
+  ).run(now, id)
+}
+
 export function getSessionCachedPrompt(id: string): {
   systemPrompt: string
   tools: import('../llm/types.js').LLMToolDefinition[]
@@ -248,8 +258,7 @@ export function getSessionCachedPrompt(id: string): {
   `,
     )
     .get(id) as
-    | { cached_system_prompt: string | null; cached_tools: string | null; cached_hash: string | null }
-    | undefined
+    { cached_system_prompt: string | null; cached_tools: string | null; cached_hash: string | null } | undefined
 
   if (!row || !row.cached_system_prompt || !row.cached_tools || !row.cached_hash) {
     return null
@@ -282,8 +291,7 @@ export function getSessionMessageCount(id: string): number {
   try {
     const db = getDatabase()
     const row = db.prepare(`SELECT message_count FROM sessions WHERE id = ?`).get(id) as
-      | { message_count: number }
-      | undefined
+      { message_count: number } | undefined
     return row?.message_count ?? 0
   } catch {
     // Database not initialized (test scenarios) - return 0
--- a/src/server/session/manager.execution-context.test.ts
+++ b/src/server/session/manager.execution-context.test.ts
@@ -64,8 +64,16 @@ const mockProviderManager = {
 import { loadConfig } from '../config.js'
 import { closeDatabase, getDatabase, initDatabase } from '../db/index.js'
 import { createProject } from '../db/projects.js'
+import { updateSessionBranch } from '../db/sessions.js'
 import { initEventStore } from '../events/index.js'
 import { SessionManager } from './manager.js'
+
+async function setSessionBranch(manager: SessionManager, sessionId: string, branch: string) {
+  updateSessionBranch(sessionId, branch)
+  // Force the manager to reload from DB so session.branch reflects the write
+  await new Promise<void>((resolve) => setImmediate(resolve))
+  expect(manager.getSession(sessionId)?.branch).toBe(branch)
+}
 
 describe('SessionManager.switchWorkspace – execution context integrity (issue #190)', () => {
   let workdir: string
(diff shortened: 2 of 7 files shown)


co-l:
Hello,

It's merged but in the review I missed that part :

> New SessionManager.clearCachedPrompt() (and the matching
> clearSessionCachedPrompt DB helper) invalidates the cached system
> prompt, resets the warmup flag and sets dynamicContextChanged = true.
> SessionManager.switchWorkspace() calls SessionManager.clearCachedPrompt()
> after the new workdir/branch has been read authoritatively and
> only on a successful mutation — no fake refreshed context on failure.

The cost of this is too high. For example, I can do a whole PR review in a workspace, with instructions at the end to leave the workspace and clean it up. If your change was live right now, I'd be paying full prompt processing tax for my last step.

This what I'm fighting for: stable cache. This is paramount for people with 200 tps prompt processing speed (less so for my setup with 2000 tps).

But I understand what you wanted to achieve ; more model awareness of current context, and system prompt is fighting against that.

So the fix I introduced is along those lines:

> Restores the "cache is sacred" invariant: `switchWorkspace` no longer invalidates the cached system prompt — the new workspace/branch is conveyed via the injected `<system-reminder>` instead, and the base prompt now tells the model to trust the latest reminder over the static `Working directory` line.

theshwal:
Thanks for the clarification. I understand the concern and agree that preserving the stable prompt cache is the right constraint, especially for slower local models.

I am going to revisit the problem with that invariant in mind:

«The cache is sacred.»

The goal will be to find a way to keep the cached system prompt untouched while still ensuring that the model reliably sees the authoritative workspace, path and branch after a mutation, including during a same-turn continuation.

I will first analyse the current reminder-based flow and its test coverage. At minimum, I will try to add an end-to-end regression test proving that:

- the same cached system prompt is reused;
- the updated workspace context reaches the next LLM call;
- failed mutations never expose a false refreshed context.

If the current implementation already satisfies this contract, the contribution may be tests only. If a gap is demonstrated, I will look for the smallest possible fix that remains fully aligned with the stable-cache design.

PR #195, opened 2026-08-01
Allow an API key in the manual vision fallback config

fullya99:
The vision fallback can already authenticate against a hosted endpoint. `describeImage` sets the header as soon as the backend is openai and a key is present:

```ts
if (isOpenAI && visionModel.apiKey) {
  headers['Authorization'] = `Bearer ${visionModel.apiKey}`
}
```

The catch is that nothing can put a key there unless the vision model comes from a configured provider. `resolveVisionFallback` only forwards `apiKey` in the `providerModelRef` branch. The manual branch returns url, model, timeout and backend, and that is it.

That manual branch is exactly where the onboarding form lands when no provider exposes a vision capable model, which is the common case when your main provider is text only. So picking "OpenAI-compatible" and pointing it at a hosted API sends the request with no Authorization header and gets a 401 back. The form has no field for the key either, and hand editing config.json does not help because `visionFallbackSchema` is a plain `z.object` and drops the field on load.

What changed:

- `apiKey` is an optional field on `visionFallbackSchema`
- `resolveVisionFallback` forwards it in the manual branch
- the PUT and test routes for `/api/config/vision-fallback` accept it
- the onboarding vision step shows a password input for it, only when the backend is openai, since ollama has no use for it

Leaving the field empty gives exactly the current behavior, and `providerModelRef` still takes precedence when it resolves, so a provider key is never shadowed by a manual one. There is a test for each of those two points.

One small refactor came along with it. `VisionModelConfigSectionProps` now extends `ManualVisionConfigProps` instead of restating its fields, because one more field pushed the two blocks past the jscpd threshold. It also means the section no longer threads each field through by hand.

Checked with `npm run typecheck`, `npm run lint` and `npm run duplicate`, all green, and `src/cli/config.test.ts` passes with the two added cases. Note that `npm run format` and a few e2e setup hooks already fail on a clean main in my environment, neither is touched here.

### DIFF
--- a/src/cli/config.test.ts
+++ b/src/cli/config.test.ts
@@ -499,6 +499,41 @@ describe('resolveVisionFallback', () => {
     expect(result!.apiKey).toBeUndefined()
   })
 
+  it('passes the api key through when the manual config sets one', () => {
+    const config = makeConfig({
+      visionFallback: {
+        enabled: true,
+        url: 'https://api.openai.com/v1',
+        model: 'gpt-4o-mini',
+        timeout: 120,
+        backend: 'openai' as const,
+        apiKey: 'sk-manual-key',
+      },
+    })
+    const result = resolveVisionFallback(config)
+    expect(result).toBeDefined()
+    expect(result!.baseUrl).toBe('https://api.openai.com/v1')
+    expect(result!.backend).toBe('openai')
+    expect(result!.apiKey).toBe('sk-manual-key')
+  })
+
+  it('keeps using the provider key when providerModelRef resolves', () => {
+    const config = makeConfig({
+      visionFallback: {
+        enabled: true,
+        providerModelRef: 'vision-provider/gpt-4o-vision',
+        url: 'https://api.openai.com/v1',
+        model: 'gpt-4o-mini',
+        timeout: 120,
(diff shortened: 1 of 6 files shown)


fullya99:
This landed before I got to the review notes, but all three of them hold and they are in `develop` now. Writing them down in case a follow up is wanted.

The onboarding one is the one with teeth. `/api/config` does not return `visionFallback.apiKey`, so the field in the vision step always initializes empty, and `handleFinish` sends `apiKey` whenever the backend is openai. The route it reaches is `POST /api/init/config` rather than the vision fallback PUT, and that merge is a plain spread with no filtering at all, so rerunning onboarding overwrites a stored key with an empty string.

Fixing it by having `/api/config` return the key would mean publishing a secret over a GET the UI reads, so the better direction is to leave the field out of the payload when it is blank.

The settings PUT keeps an empty string as though it were a key, where the provider route reads the same input as a request to forget it.

And a key made only of spaces stays truthy all the way to `vision-fallback.ts`, so it goes out as a `Bearer` header with nothing after it.

I have the three of them on a branch with tests. Happy to open it as a PR if that is useful, or to leave it alone if you would rather fold them in yourself.

PR #197, opened 2026-08-01
Connect MCP servers that require OAuth

fullya99:
OpenFox can only authenticate to an HTTP MCP server with something static. The transport is built with headers and nothing else:

```ts
const httpOpts: StreamableHTTPClientTransportOptions = {}
if (entry.config.headers) {
  httpOpts.requestInit = { headers: entry.config.headers }
}
```

That covers a server that hands out API keys. It does not cover a server that expects a sign in, which is most of the hosted ones now, and it does not cover anything sitting behind an identity aware proxy. There is no value you can type into the headers field that gets you in, so those servers are simply unreachable today.

The SDK already carries the client side of this. `StreamableHTTPClientTransport` takes an `authProvider`, and once it has one, a 401 drives protected resource discovery, dynamic client registration, the code exchange and the refresh on its own. So this is mostly a provider implementation and the plumbing to run the browser round trip, not an OAuth implementation.

What changed:

- `oauth` is an optional boolean on http MCP servers, with a checkbox in the add and edit form. Leaving it off is exactly the current behavior, `authProvider` is never set and the headers are passed through untouched
- `McpOAuthProvider` implements `OAuthClientProvider`. It records the authorization URL rather than navigating to it, since it runs on the server, and the route hands that URL back to the UI
- credentials go to `mcp-auth.json` beside `config.json`, written 0600 through a temporary file and a rename. Nothing lands in `config.json`, which is meant to be shareable, and no token is logged or returned by an API route
- four routes to start an authorization, receive the callback, paste a callback back by hand, and forget the stored credentials

On the redirect URI. The default is the loopback address the server already listens on, so a normal install needs no configuration, and authorization servers stay happy because native apps have made loopback redirects routine. When OpenFox is reached through a tunnel, that address means nothing to the browser and the tab lands on a page that will not load. The address bar still holds the code, and pasting the whole thing back finishes the flow. `OPENFOX_PUBLIC_URL` is there for anyone who would rather register a public redirect URI and keep it automatic.

Two things worth a closer look in review:

The callback route is added to `publicPaths`. It is a redirect coming from an authorization server, so it cannot present the `x-session-token` header, and without this it answers 401 in exactly the deployment that needs it. What guards it is the `state`, which is 256 bits, single use, and required on the paste path as well.

A static `Authorization` header is dropped for servers that have OAuth on. `_commonHeaders` in the SDK merges `requestInit` headers after the ones it derives from the provider, so a leftover header would shadow the token and every request would come back unauthorized with nothing on screen to explain it.

Tests cover the transport wiring both ways, the header stripping, the patch semantics on update, the store permissions including the case where the file already existed at 0644, the single use state, concurrent writes, and a corrupted store.

Checked with `npm run typecheck`, `npm run lint` and `npm run duplicate`, all green. `npm run format` and two assertions in `src/server/routes/workspace-config.test.ts` already fail on a clean main in my environment, neither is touched here.

### AI-Enhanced Development

- **AI Models:** Claude Opus 5

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@ _Vision Fallback — Image processing configuration_
 | `OPENFOX_LLM_IDLE_TIMEOUT`           | `300000`                   | LLM stream idle timeout in ms                                   |
 | `OPENFOX_PORT`                       | `10369`                    | Server listen port                                              |
 | `OPENFOX_HOST`                       | `127.0.0.1`                | Server bind host                                                |
+| `OPENFOX_PUBLIC_URL`                 | —                          | Public origin used to build the MCP OAuth redirect URI          |
 | `OPENFOX_WORKDIR`                    | `cwd`                      | Working directory                                               |
 | `OPENFOX_DB_PATH`                    | `./openfox.db`             | SQLite database path                                            |
 | `OPENFOX_LOG_LEVEL`                  | `info`                     | Log level: `debug`, `info`, `warn`, `error`                     |
--- a/src/cli/config.ts
+++ b/src/cli/config.ts
@@ -127,6 +127,7 @@ const mcpServerSchema = z.object({
   env: z.record(z.string(), z.string()).optional(),
   url: z.string().optional(),
   headers: z.record(z.string(), z.string()).optional(),
+  oauth: z.boolean().optional(),
   disabledTools: z.array(z.string()).optional(),
   cachedTools: z.array(cachedToolSchema).optional(),
   timeout: z.number().positive().optional(),
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -27,6 +27,8 @@ import {
   setNotifyMcpServersChanged,
 } from './tools/mcp-config.js'
 import { getSessionDisabledServers, setSessionDisabledServers } from './mcp/session-overrides.js'
+import { setMcpOAuthStoreMode, setMcpOAuthStorePath } from './mcp/oauth-store.js'
+import { setMcpOAuthServerPort } from './mcp/oauth-provider.js'
 import { createServerMessage } from '../shared/protocol.js'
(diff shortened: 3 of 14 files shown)


fullya99:
Both points are addressed, with a correction on each.

On the completion route. The check was genuinely weaker than the one on `/start`, and it now reads the live config through a small helper. The reachable case is narrower than described though. A `state` only ever enters the store through `McpOAuthProvider.state()`, which runs from `/start` or from `connectServer`, and both of those already require `transport === 'http'` and `oauth === true`. So the state match was implicitly re-deriving the gate, but only as of the moment the state was written. What actually slips through is a server switched to `oauth: false` between the start and the completion, since nothing clears the pending state on update. Worth closing, and closed.

While in there I gave the same treatment to `/api/mcp/oauth/callback`, which had no config check at all, not even that the server still existed. It resolves purely by `state`, and it is the route that sits in `publicPaths`, so it was the more exposed of the two.

On the store. `readStore` did trust every entry, and an entry is now dropped unless it is an object with a string `serverUrl`. One correction there: `readMcpOAuthEntry` never threw on a null entry, because `if (!entry || entry.serverUrl !== serverUrl)` short circuits before the property access. Only `findMcpOAuthEntryByState` threw, since its loop reads `entry.state` with no guard, and that is the one the callback route uses.

Filtering on read has a consequence worth naming out loud. Every save and clear reads the store then writes the whole thing back, so a dropped entry leaves the file for good the next time any other server is touched. There is now a `logger.warn` per dropped entry and a test that pins the on disk behavior down, so it is visible instead of silent.

Two more tests, 60 in `src/server/mcp/` now.

On the checks. The body was missing the model disclosure, that part is my mistake and it is fixed. The other four are showing a cancelled run, they died together inside `npm ci` on a runner shutdown and never reached the commands they were meant to run. They pass locally. I have no rerun rights on this repo.

PR #230, opened 2026-08-08
fix(e2e): run server tests sequentially

### LINKED ISSUE #225: E2E suite times out with default local worker count but passes with maxWorkers=1
## Problem

On a clean checkout of the current `co-l/openfox:develop` branch, the E2E suite is unreliable with the default local configuration.

The default E2E configuration uses:

```ts
hookTimeout: 10_000 * CI_MULTIPLIER,
pool: 'forks',
maxWorkers: process.env['CI'] === 'true' ? 1 : 12,
```

On a local developer run (`CI !== true`), this means up to 12 E2E workers can start isolated OpenFox test servers in parallel, while each `beforeAll()` hook is limited to 10 seconds.

Multiple E2E suites then fail because:

```ts
beforeAll(async () => {
  server = await createTestServer()
})
```

does not complete before the 10-second hook timeout.

The same clean checkout and the same machine pass the E2E suite when limited to a single worker.

This was reproduced independently of any feature branch changes.

## Reproduction

Environment:

- Clean checkout of `co-l/openfox:develop`
- No unrelated feature changes applied
- Linux developer machine
- Fresh build before running the E2E suite

Default E2E run:

```bash
npx vitest run \
  --config e2e/vitest.config.ts \
  --reporter=verbose
```

Observed result:

```text
Test Files  31 failed | 12 passed | 3 skipped (46)
Tests       82 passed | 282 skipped (364)
Duration    58.51s
```

Representative failures:

```text
FAIL  e2e/auto-compaction-trigger.test.ts > Auto-Compaction Trigger
Error: Hook timed out in 10000ms.
If this is a long-running hook, pass a timeout value as the last argument or configure it globally with "hookTimeout".

beforeAll(async () => {
  server = await createTestServer({ maxContext: 100 })
})
```

```text
FAIL  e2e/auto-compaction.test.ts > Auto-Compaction
Error: Hook timed out in 10000ms.

beforeAll(async () => {
  server = await createTestServer()
})
```

```text
FAIL  e2e/builder.test.ts > Builder Mode
Error: Hook timed out in 10000ms.

beforeAll(async () => {
  server = await createTestServer()
})
```

Many suites also subsequently report errors such as:

```text
TypeError: Cannot read properties of undefined (reading 'close')
```

for:

```ts
afterAll(async () => {
  await server.close()
})
```

These appear to be secondary failures: `server` was never initialized because the preceding `beforeAll(createTestServer())` timed out.

## Control test

On the exact same clean checkout and machine:

```bash
rm -rf e2e/.openfox-test

npx vitest run \
  --config e2e/vitest.config.ts \
  --reporter=dot \
  --maxWorkers=1
```

Observed result:

```text
Test Files  43 passed | 3 skipped (46)
Tests       315 passed | 49 skipped (364)
Duration    439.19s
```

No E2E test files failed with `maxWorkers=1`.

## Comparison

Same repository state, same machine, same E2E suite:

```text
default local configuration (maxWorkers=12)
→ 31 failed / 12 passed / 3 skipped

maxWorkers=1
→ 43 passed / 3 skipped
→ 315 tests passed
→ 0 failed
```

This strongly suggests that the current local E2E concurrency combined with the 10-second hook timeout can overload some developer environments enough for `createTestServer()` startup to miss the timeout.

## Why this matters

The current pre-commit hook runs:

```text
lint
typecheck
format
duplicate
test
```

in parallel, and `test` runs the unit + E2E suite.

The E2E problem already reproduces when running the E2E suite by itself, so the pre-commit hook is not the sole cause.

However, because the full test suite is part of the mandatory pre-commit gate, this E2E instability can prevent otherwise valid commits on a developer machine.

## Expected behavior

The default local E2E configuration should complete reliably on a supported developer environment.

A contributor should not need to retry commits until the E2E scheduler happens to stay below the timeout threshold.

## Possible directions

Not prescribing a particular implementation, but possible areas to investigate include:

- adapt or reduce local E2E worker concurrency;
- increase or derive `hookTimeout` according to concurrency or available resources;
- make `createTestServer()` startup more robust under parallel load;
- avoid combining the full E2E suite with other resource-intensive checks in the pre-commit path;
- use a different worker policy for local development and CI.

The important observation is that the current A/B result is deterministic on this machine:

```text
maxWorkers=12 → widespread createTestServer() hook timeouts
maxWorkers=1  → full E2E pass
```

This reproduction was obtained from a clean upstream checkout, without the unrelated feature changes that originally exposed the issue.

SulimanAbdulrazzaq:
### Summary

Closes #225.

Each E2E test file starts a full OpenFox server. Starting up to 12 of these servers concurrently can exhaust local CPU and memory long enough for `createTestServer()` to exceed the existing 10-second hook timeout.

Run server-backed E2E files sequentially, matching the already reliable CI worker policy. This addresses the resource contention directly without increasing test or hook timeouts.

A regression assertion now protects the worker policy.

Validation:

- Reproduced the default-worker startup timeout cascade
- Confirmed four workers still fail with the same hook timeouts
- `e2e/timeouts.test.ts`: 4 tests passed
- E2E TypeScript check passed
- Prettier passed for both changed files
- Full sequential E2E run was capped after approximately eight minutes under concurrent batch load and did not produce a final result

### AI-Enhanced Development

- **AI Models:** GPT-5.6 (Codex)

### Cache Impact

- **No**

### DIFF
--- a/e2e/timeouts.test.ts
+++ b/e2e/timeouts.test.ts
@@ -11,6 +11,10 @@ describe('E2E timeouts', () => {
     expect(config.test?.testTimeout).toBe(15_000 * CI_MULTIPLIER)
   })
 
+  it('runs server-backed test files sequentially', () => {
+    expect(config.test?.maxWorkers).toBe(1)
+  })
+
   it('keeps websocket waits within the per-test budget', () => {
     const budget = 8_000 * CI_MULTIPLIER
     expect(DEFAULT_WAIT_TIMEOUT_MS).toBeLessThanOrEqual(budget)
--- a/e2e/vitest.config.ts
+++ b/e2e/vitest.config.ts
@@ -14,10 +14,10 @@ export default defineConfig({
     testTimeout: 15_000 * CI_MULTIPLIER,
     hookTimeout: 10_000 * CI_MULTIPLIER,
 
-    // Run tests in parallel with fork pool
-    // Each test file gets its own in-process server on a dynamic port
+    // Each test file starts a full server. Parallel startups contend for enough
+    // CPU and memory to exceed the hook timeout on supported developer machines.
     pool: 'forks',
-    maxWorkers: process.env['CI'] === 'true' ? 1 : 12, // CI runners are resource-constrained
+    maxWorkers: 1,
 
     // No global setup - each test file manages its own server
     // globalSetup: './setup.ts',  // REMOVED - using in-process servers


co-l:
Thanks for digging into this — the diagnosis (parallel full-server startups blowing the hook budget) is spot on.

We went with a gentler fix on `develop` instead of forcing sequential for everyone: the worker count is now configurable per machine via an env var, with the old defaults preserved.

- Default: 12 workers locally, 1 on CI (unchanged).
- Override per machine: `OPENFOX_E2E_MAX_WORKERS=<n>`
  - e.g. `OPENFOX_E2E_MAX_WORKERS=4 npm run test:e2e`
- Invalid values (non-integer, <1, garbage) fall back to the default.

The concern with merging this PR as-is was that hardcoding `maxWorkers: 1` — plus the new regression test asserting it — would permanently remove the ability to parallelize for everyone, including machines that handle 12 fine. This way the knob stays.

Closing this one; if your machine still can't cope with the env var set low, reopen and we can look at a smarter default (e.g. derived from CPU count).

Issue #220, opened 2026-08-07
bug: Mistral API returns HTTP 422 when thinking is enabled (reasoning field rejected in message history)


SamirSaidani:
When using the Mistral API with thinking enabled, OpenFox sends back the `reasoning` field in the conversation history (previous assistant messages). Mistral rejects this with HTTP 422:

```
{"detail":[{"type":"extra_forbidden","loc":["body","messages",N,"assistant","reasoning"],
"msg":"Extra inputs are not permitted"}]}
```

**Root cause:** Mistral supports `reasoning_effort` to request thinking on a given turn, but does not accept the `reasoning` field in outgoing assistant messages (i.e. conversation history). OpenFox already has a `sendReasoningInMessages` provider-level flag to handle exactly this case, but it is not auto-detected for Mistral.

**Workaround:** Manually uncheck "Send reasoning in messages" in Provider-Level Defaults (⚙️ icon in the provider settings).

**Proposed fix (PR ready):**
- Add a `probeReasoningInMessages` probe to the auto-config flow: after detecting thinking support, send a test request with a `reasoning` field in an assistant message. If the API rejects it (non-2xx), set `sendReasoningInMessages: false` automatically.
- Apply the result in `ProviderModal` so the checkbox is auto-unchecked when saving the provider.

This keeps the two concerns separate: `reasoning_effort` (thinking level per request) is unaffected — only the echoing of past reasoning in message history is disabled.

**Broader value:** Beyond Mistral, this probe would benefit any provider that supports thinking but rejects extra fields in message history. Rather than users having to discover `sendReasoningInMessages` manually, auto-config would detect and configure it transparently for any such API.

PR #237, opened 2026-08-12
feat: per-model control of which sampling params are sent to the backend

ikarys:
### Summary

Some backends reject sampling parameters they don't support (e.g. `temperature`, `top_p`, `top_k`, `reasoning_effort`), returning HTTP 400 and blocking the conversation. OpenFox already stripped these via an `omitParams` field set by auto-config, but users had no manual control over which params are sent.

This PR adds a **"Send" checkbox** next to each sampling parameter in the provider model editor (Advanced section), so users can explicitly enable/disable individual params per model. Unchecking a param adds it to `omitParams` and strips it from outgoing requests; re-checking removes it.

#### What changed

- **ProviderModal**: New `SamplingParamField` component with a "Send" checkbox per param (Temperature, Top P, Top K, Max tokens). Unchecking disables + dims the input and adds the param key to `omitParams`. Re-checking re-enables it. Pre-existing `omitParams` entries (e.g. `reasoning_effort` from auto-config) are preserved on toggle.
- **ProviderModal**: Conditional "Re-enable reasoning_effort" checkbox in the Thinking section, shown only when auto-config omitted it. Lets users re-enable `reasoning_effort` without re-running auto-config.
- **ProviderModal**: Omitted fields now show a "Not sent" placeholder instead of a stale dimmed value, making it clear the value won't be sent.
- **Refactor**: Collapsed 4 near-identical sampling-param blocks (~80 lines) into one parameterized `SamplingParamField` component. Net LOC reduction.
- **Comments**: Genericized all vendor-specific comments and test fixtures to vendor-neutral wording.

#### How it works

```
omitParams: ['temperature', 'top_p']  →  temperature and top_p stripped from request body
omitParams: undefined                  →  all params sent (default)
```

The `omitParams` field was already wired through the full stack (ModelConfig → provider-manager → client-pure → request body). This PR exposes it in the UI so users can control it manually, without relying solely on auto-config detection.

#### Related issues

- **#220** — A backend returns HTTP 422 when thinking is enabled (a field rejected in message history). Same class of problem: a backend rejecting a field OpenFox sends. This PR gives users manual control over which params are sent, addressing the broader pattern. (Note: #220 is specifically about `reasoning` in message history, handled by the existing `sendReasoningInMessages` flag — but the underlying need for per-field control is the same.)

### AI-Enhanced Development

- **AI Models:** glm-5.2, claude-opus-5

### Cache Impact

- **No** — `omitParams` strips generation-time parameters (`temperature`, `top_p`, `top_k`, `max_tokens`, `reasoning_effort`) from the request body. These are not part of the cached prompt prefix (system prompt + tools + messages) or the tool fingerprint. The dynamic context hash and KV cache are unaffected.

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm run lint` — clean
- [x] `npx prettier --check` (modified files) — clean
- [x] `npx vitest run web/src/components/shared/ProviderModal.test.tsx` — 18 passed
- [x] `npx vitest run src/server/llm/client-pure.test.ts src/server/providers/auto-config.test.ts src/server/provider-manager.test.ts` — 75 passed
- [x] New tests: Send checkboxes rendered + checked by default, uncheck → omitParams + dim + clear, recheck → removed, preserve pre-existing entries, auto-config omitParams reflected, reasoning_effort re-enable

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,7 @@ e2e/.openfox/
 e2e-playwright/.openfox/
 e2e/.openfox-test/
 e2e/**/.openfox-test/
+.openfox/pr-description.md
 
 # Logs
 *.log
--- a/src/server/chat/agent-loop.ts
+++ b/src/server/chat/agent-loop.ts
@@ -127,6 +127,7 @@ export interface TopLevelLoopConfig {
     supportsVision?: boolean
     chatTemplateKwargs?: Record<string, unknown>
     queryParams?: Record<string, unknown>
+    omitParams?: string[]
   }
   signal?: AbortSignal | undefined
   onMessage?: ((msg: ServerMessage) => void) | undefined
--- a/src/server/chat/stream-pure.test.ts
+++ b/src/server/chat/stream-pure.test.ts
@@ -102,6 +102,36 @@ describe('stream-pure', () => {
     })
   })
 
+  it('excludes omitted params from result.modelParams so stats reflect the wire request', async () => {
+    const client = createMockClient([
+      { type: 'text_delta', content: 'hi' },
+      {
+        type: 'done',
+        response: {
+          id: 'resp-omit',
+          content: 'hi',
+          toolCalls: [],
+          finishReason: 'stop',
+          usage: { promptTokens: 5, completionTokens: 5, totalTokens: 10 },
+        },
+      },
+    ])
+
+    const gen = streamLLMPure({
+      messageId: 'msg-omit',
+      systemPrompt: 'system',
+      llmClient: client,
+      messages: [{ role: 'user', content: 'hello' }],
+      modelSettings: { omitParams: ['temperature', 'max_tokens'] },
+    })
+
+    const result = await consumeStreamGenerator(gen, () => {})
+
+    expect(result.modelParams).not.toHaveProperty('temperature')
+    expect(result.modelParams).not.toHaveProperty('maxTokens')
+    expect(result.modelParams).toHaveProperty('topP')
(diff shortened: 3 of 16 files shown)


co-l:
Merged, will be released in v2.0.119

AI-summary

Adds per-model control over which sampling parameters are sent to the LLM backend. Auto-config now probes each model to detect params the backend rejects (with a control-probe guard added during review to avoid false attribution), and the builder strips them from outgoing requests. The UI exposes per-param "Send" checkboxes plus a recovery control for `reasoning_effort`.

---

- **`omitParams` on `ModelConfig`** — strip rejected top-level sampling params (`temperature`, `top_p`, `max_tokens`, `top_k`, `reasoning_effort`) from outgoing requests; runs after all merges so it wins over `queryParams` additions
- **Auto-detection** — auto-config probes each model with an agentic-loop-shaped request (incl. a dummy tool) and reports params rejected via HTTP 400; a control probe skips detection when the backend rejects the probe structurally, and word-boundary matching avoids false positives (e.g. `max_tokens_budget` ≠ `max_tokens`)
- **Accurate stats/retries** — `modelParams` (turn stats + truncation-retry budget) now reflects the actual wire request, so omitted params are neither sent nor reported as sent
- **UI** — per-sampling-param "Send" checkboxes with disabled inputs, a "Re-enable reasoning_effort" recovery checkbox, and auto-config results pre-populating the checkboxes
- **Plumbing** — threaded through provider config, provider-manager, the providers API, and the streaming request path

PR #251, opened 2026-08-18
fix(web): scope session-list reloads to the active project after mutations

ikarys:
### Summary

Deleting a session from a project sidebar shrank the list far below the project's real count (e.g. 10 -> 3) and a page refresh restored it.

**Root cause:** post-mutation reloads (\`deleteSession\`, \`renameSession\`, \`toggleFavorite\`, \`deleteAllSessions\` and the \`session.deleted\` / \`session.deletedAll\` WS handlers) called \`listSessions()\` with no \`projectId\`, fetching the 20 most-recent sessions across ALL projects. \`mergeSessionSummaries\` then replaced the whole \`sessions\` array with that global subset, and the sidebar's per-project filter dropped the other projects' sessions.

**Race condition found while fixing:** the server broadcasts \`session.deleted\` BEFORE responding to the HTTP DELETE, so the WS handler wiped the session from state before \`deleteSession\` resolved its \`projectId\` -> \`undefined\` -> global reload. The fix resolves \`projectId\` BEFORE the network request.

**Fix (web-only):**
- \`mergeSessionSummaries\` is project-aware: a scoped reload REPLACES that project's slice while preserving other projects; the home/global path REPLACES entirely (preserves \`listHomeSessions\` curated-set behavior).
- \`deleteSession\`, \`renameSession\`, \`toggleFavorite\` resolve the session's \`projectId\` before the request and reload scoped.
- \`deleteAllSessions\` reloads scoped to its \`projectId\` param.
- WS \`session.deleted\` removes the session from \`state.sessions\` immediately and reloads scoped; \`session.deletedAll\` removes the project's sessions immediately and reloads scoped to \`message.sessionId\` (the projectId).
- \`loadMoreSessions\` offsets by the target project's loaded count, not the global \`sessions.length\` (which now spans multiple projects).
- \`resolveSessionProjectId\` is shared via \`panes.ts\` (no import cycle).

**Tests:** 8 new Vitest cases covering cross-project delete/rename/favorite/deleteAll, the WS race condition, \`listHomeSessions\` replace semantics, and \`loadMoreSessions\` offset.

### AI-Enhanced Development

- **AI Models:** GLM 5.2

### Cache Impact

- **No**

### DIFF
--- a/web/src/stores/session/messageHandler.test.ts
+++ b/web/src/stores/session/messageHandler.test.ts
@@ -559,7 +559,6 @@ describe('chat.stats handler', () => {
     generationTokens: 10,
     generationSpeed: 5,
   } as any
-
   beforeEach(() => {
     wsSendMock.mockClear()
     wsSubscribeMock.mockClear()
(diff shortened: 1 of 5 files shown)


co-l:
---
### AI-summary
- Rebased onto latest develop and resolved the merge conflict in the message-handler tests — kept both the new session.deleted/session.deletedAll suites and the existing chat.stats suite
- Verified the full affected suite (90 tests green, typecheck and lint clean)
- Pushed the resolved branch to the fork so the PR merges cleanly with full attribution

PR #255, opened 2026-08-19
feat(workflows): per-step model overrides, team presets & llm_decision dynamic orchestration

schpill:
## Summary

This adds a composable, plugin-friendly way to run **multi-model workflows where an LLM decides the sequence** — "which model orchestrates, which produces code, which verifies/corrects" — configurable per **step** with no fixed role cap. It is fully backward-compatible (zero changes to existing workflows or APIs) and built on top of the existing plugin + transition-condition infrastructure.

The motivating use case: a team wants a workflow where a cheap model orchestrates, a strong model writes code, and a verifier model checks + routes back to the builder or forward to review. Today you can pin one model per *agent*, but you cannot pin a model per *step*, you cannot group agents into a reusable team preset, and no transition can ask an LLM "what runs next?". This PR closes all three gaps.

### Changes

Five commits, each independently testable, layered on the previous:

1. **`feat(workflows): pluggable custom transition handlers via plugin API`** — Plugins can now register custom `TransitionHandler`s through `ProviderPluginRegistry.registerTransitionHandler(name, handler)`. Workflows reference them via `when: { type: 'custom', handler: '<name>', config: {...} }`. The plugin discovery path (`configDir/plugins` + `node_modules`) already existed; this just opens the registry to transition logic. A plugin can override built-ins.

2. **`feat(agents): per-step model override (workflowId:stepId)`** — `resolveLLMClientForStep(workflowId, stepId, agentId, fallback, providerManager)` with precedence **step override > team assignment > agent override > session**. Overrides stored in DB settings, keyed `workflowId:stepId`. So N steps → N independently pinnable models. No 4-role (or any) cap — it is pure key/value.

3. **`feat(teams): workflow team presets (N roles, arbitrary models)`** — Team = named set of `{ agentId → { providerId, model, reasoningEffort? } }` assignments, stored in DB settings. Bind a team to a workflow (`PUT /api/teams/bindings/:workflowId { teamId }`). A step whose agent is in the bound team resolves to the team's model. Teams are reusable across workflows and hold an arbitrary number of roles.

4. **`feat(workflows): add built-in llm_decision transition handler`** — A built-in `llm_decision` transition handler (registered before plugins load, so plugins can override it). A step whose transitions carry `when: { type: 'custom', handler: 'llm_decision', config }` asks the orchestrator LLM to pick the next step. Each sibling transition declares the shared `candidates` list plus its own `thisGoto`; the handler makes **one LLM call per (workflow, step, outcome)** and fires only the transition whose `thisGoto` matches the choice. The orchestrator LLM is the step's *resolved* client — so "which LLM orchestrates" is itself configurable per step via the same override/team machinery. On no client, call failure, or unparseable response it returns `false` so a following `always` fallback wins — the workflow never blocks on a routing decision.

5. **`feat(workflows): Phase 4 — dream orchestration demo + validation + step override routes`** — HTTP surface (`GET/PUT/DELETE /api/workflows/:workflowId/steps/:stepId/model`) and team routes (`/api/teams`). A loadable demo workflow (`.openfox/workflows/dream.workflow.json`) routes build/verify/review via `llm_decision`. `docs/DREAM-ORCHESTRATION.md` documents setup, resolution precedence, and team binding. A shared `readOverrideFields` helper keeps the step/agent override routes duplicate-free (jscpd 0 clones).

### Implementation notes

- **No API removals or renames.** All additions are additive. Existing workflows, agents, and endpoints behave identically.
- **Resolution precedence** (highest first): step override → workflow team assignment → agent override → session client. Documented in `docs/DREAM-ORCHESTRATION.md` and enforced in `resolveLLMClientForStep`.
- **Dynamic orchestration, not 4 separate agents.** The `llm_decision` handler is what makes this a *team*: a single LLM call per step outcome inspects the step's result and chooses the next step from the declared candidates. Sibling transitions share that one call via a module-level cache keyed by `(workflowId, stepId, outcome, candidates)`, so adding more candidate routes does not add more LLM calls.
- **Graceful degradation.** Missing client / rejected call / unparseable reply → handler returns `false` → an `always` transition (or the next matching one) wins. The workflow never deadlocks on routing.
- **Step context flows to the client factory.** `sessionManager.createClientForAgent` now receives `{ workflowId, stepId }` so a step-resolved (team) client — not the session client — is what `llm_decision` consults. This is verified by an integration test (`phase4-integration.test.ts`): routing follows the LLM's choice using the *team* client, and the session client is never called.

### Testing

All developed TDD. Full unit suite passes (**4411 / 4411**), typecheck + lint green, jscpd 0 clones.

New tests:
- `llm-decision-handler.test.ts` (9) — fires only the matching `thisGoto` with one call; caches across siblings; re-queries per outcome; parses verbose responses (substring fallback); unparseable → all false; no client → false; rejected call → false; forwards `signal` + `temperature`; prompt includes outcome + candidate labels.
- `phase4-integration.test.ts` (2) — proves composition: the step-resolved (team) client is what `llm_decision` consults (session client NOT consulted); routing follows the LLM choice (verifier → reviewer, not builder); unparseable → always fallback.
- `step-model-overrides.test.ts` (7) — null when unset; set + read; invalid `reasoningEffort` → 400; partial body → 400; empty body clears; delete; per-step independence.
- `teams` routes + registry tests for presets and workflow bindings.
- Existing executor tests (`executor-mode`, `executor-retry`, `executor-resume-nudge`, `executor-subgroup`) updated for the new `createClientForAgent` signature.

## AI-Enhanced Development

- **AI Models:** GLM 5.2 via OpenFox agent

## Cache Impact

- **No** — no changes to system prompts, tool definitions, skills, or cached context. This is server-side workflow/agent/route logic only.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,5 @@ tmp/
 worktrees/
 
 plugins/
+# jscpd JSON report (generated, .md is the tracked stub)
+report/jscpd-report.json
--- a/.openfox/agents/dream-reviewer.agent.md
+++ b/.openfox/agents/dream-reviewer.agent.md
@@ -1,0 +1,19 @@
+---
+id: dream-reviewer
+name: Dream Reviewer
+description: Top-level workflow agent for the dream demo's review step. Reviews the change for correctness and style, then calls step_done().
+subagent: false
+color: '#a855f7'
+allowedTools:
+  - read_file
+  - run_command
+---
+
+You are a code reviewer running as a top-level workflow step.
+
+Your job:
+
+1. Review the completed change for correctness and style.
+2. Approve it or note requested changes, concisely.
+
+When you are finished, you MUST call `step_done()` to signal completion. Do not loop: one review pass + `step_done()` is enough.
--- a/.openfox/agents/dream-verifier.agent.md
+++ b/.openfox/agents/dream-verifier.agent.md
@@ -1,0 +1,22 @@
+---
+id: dream-verifier
+name: Dream Verifier
+description: Top-level workflow agent for the dream demo's verify step. Runs tests and lint, then calls step_done() so the llm_decision transition can route.
+subagent: false
+color: '#22c55e'
+allowedTools:
+  - read_file
+  - run_command
+  - session_metadata
+  - web_fetch
+---
+
+You are a verifier running as a top-level workflow step.
+
+Your job:
+
+1. Run the project's tests and lint (if any exist).
+2. Read the changed files to confirm the work meets the request.
+3. Report the outcome concisely.
+
+When you are finished, you MUST call `step_done()` to signal completion. Do not loop: a single verification pass + `step_done()` is enough.
--- a/.openfox/workflows/dream.workflow.json
+++ b/.openfox/workflows/dream.workflow.json
@@ -20,7 +20,9 @@
       "prompt": "Implement the requested change. When you are done, call step_done().",
       "transitions": [
         {
-          "when": { "type": "always" },
+          "when": {
+            "type": "always"
+          },
           "goto": "verify"
         }
       ]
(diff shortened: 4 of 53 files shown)


JamesDAdams:
@schpill We made a big update, you need to refactor

Issue #252, opened 2026-08-19
The Qwen3.8 reasoningEfforts contains some errors.


rankaiyx:
./src/server/providers/model-catalog.ts
./src/server/providers/model-catalog.test.ts
```
  {
    // Qwen3.6 / 3.7 / 3.8 families: low/medium/high is the common denominator.
    pattern: /^qwen3\.(6|7|8)(\b|-)/i,
    entry: { reasoningEfforts: ['low', 'medium', 'high'], defaultReasoningEffort: 'medium' },
  },
```

It seems that only version 3.8 actually has this feature. This feature is not available in version 3.6, and it is not open source in version 3.7, so we have no way of knowing about it.

And it's not "high", it's "xhigh".

https://huggingface.co/Qwen/Qwen3.6-27B/blob/main/chat_template.jinja
https://huggingface.co/Qwen/Qwen3.8-27B/blob/main/chat_template.jinja
https://huggingface.co/Qwen/Qwen3.8-2.4T-A95B/blob/main/chat_template.jinja

rankaiyx:
I used the following switch in llama-server to view the final prompts rendered from the chat template.
--log-prompts-dir PATH 
 Log prompts to directory (auto-created if not present; only used for debugging, default: disabled)

co-l:
Hello, absolutely! I'm fixing it.

Meanwhile you can customize the list of available efforts in the edit provider modal:

<img width="955" height="763" alt="Image" src="https://github.com/user-attachments/assets/4a65193b-7219-4a97-bce0-18344c573fc1" />

The "Effort presets" can be manually adjusted.

rankaiyx:
> Meanwhile you can customize the list of available efforts in the edit provider modal:

I've tried this, and it seems that no matter whether I set it to low or xhigh, the chat template on the server side doesn't seem to be parsed as expected.


This is the chat template I'm using:
https://huggingface.co/froggeric/Qwen-Fixed-Chat-Templates

By using "Inline Chat Tags for Direct Effort Steering", I saw the expected system prompt injection in the prompt words (--log-prompts-dir).
That is, simply type in the dialog box: <|think_xhigh|>

rankaiyx:
Prompts logged via --log-prompts-dir with <|think_xhigh|>:
```
<|im_start|>system
Reasoning effort is set to xhigh. Please think carefully through the task, validate key assumptions, consider plausible alternatives, and prioritize correctness, consistency, and clarity in the final answer.

# Tools

[...]

PR #257, opened 2026-08-20
fix(llm): route reasoning effort via chat_template_kwargs for llama.cpp

### LINKED ISSUE #256: [Bug] Reasoning effort level has no effect on llama.cpp backend
**Problem**

For a `llamacpp` provider, changing the reasoning effort level (UI picker / `thinkingLevel`) does nothing. The thinking on/off switch works, but the level (low/medium/xhigh) never reaches the model.

**Root cause**

llama.cpp's server only reads `reasoning_effort` from `chat_template_kwargs` (it populates Jinja template variables from there). A top-level `reasoning_effort` body field is silently ignored.

OpenFox always sends the effort as a top-level body field:

- `src/server/llm/client-pure.ts` — when `thinkingQueryParams` is set, the resolved effort is merged as top-level `reasoning_effort` (lines ~282-284); `chat_template_kwargs` is only injected when no query params exist, and only for per-request effort when `supportsChatTemplateKwargs` is true.
- `src/server/llm/backend.ts` — `llamacpp` has `supportsChatTemplateKwargs: false`, so the level is never routed into `chat_template_kwargs`.

Result: the chat template always sees its default effort (medium for the froggeric Qwen3.8 template, xhigh for the official one), regardless of the selected level.

**Verification (reproducible)**

Against a llama.cpp server serving Qwen3.8-27B, asked the model to quote the "Reasoning effort is set to ..." sentence from its system prompt:

| Request | Rendered prompt |
|---|---|
| top-level `reasoning_effort: "low"` | no instruction (default medium) |
| top-level `reasoning_effort: "xhigh"` | no instruction |
| `chat_template_kwargs: {"reasoning_effort": "low"}` | "Reasoning effort is set to low..." injected |

The llama.cpp prompt dump (`/v1/chat/completions` rendered prompt) confirms OpenFox sessions carry no effort instruction even when the config says `thinkingLevel: "low"`.

**Suggested fix**

For the `llamacpp` backend, send the resolved effort inside `chat_template_kwargs.reasoning_effort` instead of (or in addition to) the top-level field. Also map `none` to `enable_thinking: false`, since the official Qwen template rejects `reasoning_effort: "none"` with a 400. The `thinkingQueryParams`/`nonThinkingQueryParams` escape hatches work but are static — they cannot express a user-selected level, so a code change is required for the level picker to function.

**Workaround (config only, fixed level)**

```json
"thinkingQueryParams": "{\"chat_template_kwargs\":{\"enable_thinking\":true,\"reasoning_effort\":\"low\"}}",
"nonThinkingQueryParams": "{\"chat_template_kwargs\":{\"enable_thinking\":false}}"
```

This fixes on/off plus one hardcoded level; switching levels in the UI still has no effect.

rankaiyx:
Fixes #256, #258
Related: #252

## Problem

Two related bugs made the reasoning effort level ineffective on the llama.cpp backend:

1. **llama.cpp ignores top-level `reasoning_effort`** (#256). llama.cpp only populates Jinja chat-template variables from `chat_template_kwargs`; a top-level `reasoning_effort` body field is silently ignored, so the selected effort level never reached the model — the chat template always used its default effort.
2. **Session LLM clients were created with the `unknown` backend** (#258). `createClientForProvider` called `createLLMClient` without the backend argument, so every session/agent/sub-agent client defaulted to the `unknown` backend and its capability table. This silently disabled any backend-specific capability for session clients (for llamacpp it neutralized fix 1; for ollama/openai it could inject unsupported params like `top_k` or `chat_template_kwargs`).

## Changes

- **`src/server/llm/backend.ts`** — new `routesEffortViaChatTemplateKwargs` capability, set only for `llamacpp`
- **`src/server/llm/client-pure.ts`** — when the capability is set, the resolved effort is layered into `chat_template_kwargs.reasoning_effort` instead of the top-level field. User-provided `chat_template_kwargs` (e.g. from `thinkingQueryParams`) take precedence for `reasoning_effort`; the rest of the query params are still merged. Effort `none` maps to `enable_thinking: false` (official Qwen templates 400 on `reasoning_effort: "none"`, froggeric templates treat `none` as "thinking off")
- **`src/server/provider-manager.ts`** — `createClientForProvider` passes `provider.backend` to `createLLMClient` so session clients get the correct capability table
- **`src/server/session/name-generator.ts`** — same backend passed in the legacy client fallback path
- **`src/server/llm/types.ts`** — widen the `ReasoningEffort` type to the full shared vocabulary (`none`/`minimal` were missing, which blocked routing `none`)

## Verification

- Unit tests: effort routing into kwargs, queryParams merge, user-kwargs precedence, `none` mapping, no-op case, streaming parity; `createClient` backend propagation (vllm + llamacpp)
- End-to-end against a live llama.cpp server (Qwen3.8, built from this branch): rendered prompt dumps show "Reasoning effort is set to low/xhigh" injected in the system prompt for the matching levels, and an empty think block with no instruction for `none` — including through a real OpenFox session (previously the instruction was absent because session clients used the `unknown` backend)
- `npm run typecheck` and `npm run lint` clean; full test suite green except 3 pre-existing timezone-dependent failures in `web/src/lib/format-date.test.ts` (reproduced on the base commit, unrelated to this change)

## Notes

- Other backends (vllm/sglang/openai/ollama/...) keep their existing top-level `reasoning_effort` behavior; the only change for them is that session clients now use their real capability table instead of `unknown`

### DIFF
--- a/src/server/llm/backend.test.ts
+++ b/src/server/llm/backend.test.ts
@@ -37,12 +37,20 @@ describe('backend', () => {
       const caps = getBackendCapabilities('llamacpp')
       expect(caps.supportsChatTemplateKwargs).toBe(false)
       expect(caps.supportsTopK).toBe(true)
+      expect(caps.routesEffortViaChatTemplateKwargs).toBe(true)
     })
 
     it('returns vllm-like capabilities for unknown', () => {
       const caps = getBackendCapabilities('unknown')
       expect(caps.supportsChatTemplateKwargs).toBe(true)
       expect(caps.supportsTopK).toBe(true)
+      expect(caps.routesEffortViaChatTemplateKwargs).toBe(false)
+    })
+
+    it('does not route effort via chat_template_kwargs for other backends', () => {
+      for (const backend of ['vllm', 'sglang', 'openai', 'anthropic', 'ollama', 'lmstudio', 'opencode-go'] as const) {
+        expect(getBackendCapabilities(backend).routesEffortViaChatTemplateKwargs).toBe(false)
+      }
     })
   })
 
--- a/src/server/llm/backend.ts
+++ b/src/server/llm/backend.ts
@@ -13,6 +13,12 @@ export interface BackendCapabilities {
   supportsTopK: boolean
   /** Whether the client uses Ollama's native /api/chat (consumes num_ctx) */
   supportsNumCtx: boolean
+  /**
+   * Whether reasoning effort must be routed through chat_template_kwargs
+   * (llama.cpp only populates Jinja template variables from kwargs; a
+   * top-level reasoning_effort body field is silently ignored).
+   */
+  routesEffortViaChatTemplateKwargs: boolean
 }
 
 const BACKEND_CAPABILITIES: Record<Backend, BackendCapabilities> = {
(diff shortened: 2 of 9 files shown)


rankaiyx:
The actual test results show that the work is as expected.

co-l:
Merged, will be released in v2.0.125
---
### AI-summary
- Rebased the PR onto latest develop and resolved merge conflicts with the LLM backend capability tables (kept both the Ollama `supportsNumCtx` and the new `routesEffortViaChatTemplateKwargs` capabilities)
- Adapted the affected test capability literals to the merged `BackendCapabilities` shape and verified the fix end to end — full unit suite (4397 tests) and e2e suite green, typecheck and lint clean
- Pushed the rebased branch to the contributor's fork so the PR merges cleanly with full attribution

PR #264, opened 2026-08-22
fix(server): enforce default session history limit

S-Gr4y:
## Summary

- apply the declared `display.maxVisibleItems` default when no setting row exists
- fall back to that default for blank, malformed, negative, or fractional values
- preserve explicit `0` as the opt-in unlimited-history value
- cover the default 300-item truncation and the `?full=true` bypass

## Why

The web UI presents 300 as the default history limit, but `getMaxVisibleItems()` returned `0` when the setting had never been saved. The session route converts `0` to `undefined`, so a fresh/default configuration serialized and sent the complete conversation before the client truncated it.

This is especially noticeable in WebKit/Safari on long sessions. In the local reproduction, an 800-message response was about 8.16 MiB, while the most recent 300 messages accounted for about 3.67 MiB. Applying the existing default server-side avoids sending roughly 4.5 MiB on that session without changing the displayed default or the explicit full-history flow.

## Compatibility

- users who explicitly set `display.maxVisibleItems` keep their chosen limit
- explicit `0` remains unlimited
- `GET /api/sessions/:id?full=true` still returns the full history

## AI disclosure

- **AI Models:** GPT-5

## Tests

- `npx vitest run src/server/db/settings.test.ts src/server/routes/sessions.test.ts` (29 passed)
- `npm run typecheck`
- `npm run lint`
- `npx prettier --check src/server/db/settings.ts src/server/db/settings.test.ts src/server/routes/sessions.test.ts`
- full unit suite: 4,461 passed, 32 skipped
- full E2E suite: 325 passed, 49 skipped, with 3 unrelated transient failures in `ask-user` / `path-security`; isolated rerun passed all 24 tests

### DIFF
--- a/src/server/db/settings.test.ts
+++ b/src/server/db/settings.test.ts
@@ -1,7 +1,7 @@
 import { beforeEach, describe, expect, it } from 'vitest'
 import { loadConfig } from '../config.js'
 import { closeDatabase, initDatabase } from './index.js'
-import { SETTINGS_KEYS, deleteSetting, getAllSettings, getSetting, setSetting } from './settings.js'
+import { SETTINGS_KEYS, deleteSetting, getAllSettings, getMaxVisibleItems, getSetting, setSetting } from './settings.js'
 
 describe('db settings', () => {
   beforeEach(() => {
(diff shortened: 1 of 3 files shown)


co-l:
---
### AI-summary
- Reviewed the fix: absent/blank/invalid settings now fall back to the declared 300 default, while explicit 0 (unlimited) is preserved across all call sites
- Verified 29 tests (new edge-case coverage for blank, NaN, negative, fractional, and explicit 0/150) plus typecheck and lint clean

PR #265, opened 2026-08-22
perf(web): paginate and virtualize long session history

S-Gr4y:
### Summary

Large sessions currently transfer and mount their complete rendered history whenever the user switches panes. Tool-heavy sessions can therefore block the browser main thread even when only the latest messages are visible.

This PR:

- loads a bottom-anchored recent page (up to 10 turns, 30 items, or 1 MiB) for normal session navigation;
- adds a cursor endpoint and incremental “Load older history” flow without splitting user turns;
- preserves scroll position while prepending older messages and prevents duplicate page requests;
- mounts only the four newest feed items initially for paginated histories, then reveals older loaded items while scrolling upward;
- leaves non-paginated feeds and the existing explicit virtualization setting unchanged;
- keeps “Open full history in a new tab” as an escape hatch.

No storage schema or persisted session data is changed. Event folding still happens server-side; this PR targets transfer size and browser rendering cost.

### Measured impact

Measured on the same copied 800-message / 8.56 MB session, with warmed switches and service workers disabled (12 switches per browser):

| Metric | Pagination only | Pagination + scoped virtualization | Change |
| --- | ---: | ---: | ---: |
| WebKit median switch | 517 ms | 225 ms | -56.5% |
| WebKit p95 | 1,071 ms | 290 ms | -72.9% |
| Chromium median switch | 347 ms | 115 ms | -66.9% |
| Chromium p95 | 597 ms | 116 ms | -80.6% |
| Heavy-session DOM size | 14,335 nodes | 2,992 nodes | -79.1% |

The recent-page response contained 27 messages / 761 KB instead of 800 messages / 8.56 MB: **-91.1% transferred payload** and **-96.6% messages**. API latency stayed around 22 ms, which confirms that browser parsing/rendering was the dominant switching cost in this case.

The WebKit timings above were collected on desktop WebKit, not on a physical iPhone. The mobile Safari benefit is expected from the same reduction in DOM size, layout/paint work, JavaScript execution, and memory pressure—costs that are particularly visible on lower-power mobile CPUs. An iPhone 15 WebKit emulation smoke test at 393×659 loaded the heavy session with 2,992–3,042 DOM nodes and no console errors.

### Validation

- `npm run test:unit` — 4,467 passed, 32 skipped
- session REST E2E — 22 passed
- focused pagination/session/feed suites — 102 passed
- `npm run typecheck`
- `npm run lint`
- `npm run build:web`
- desktop WebKit and iPhone 15 WebKit-emulation smoke tests

### AI-Enhanced Development

- **AI Models:** GPT-5

### Cache Impact

- **No**

### DIFF
--- a/e2e/session-rest.test.ts
+++ b/e2e/session-rest.test.ts
@@ -154,6 +154,49 @@ describe('Session REST API', () => {
       const data: any = await response.json()
       expect(data.error).toBe('Session not found')
     })
+
+    it('loads recent history by complete turns and pages backwards without overlap', async () => {
+      const createRes = await fetch(`${server.url}/api/sessions`, {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({ projectId, title: 'Paginated session' }),
+      })
+      const created: any = await createRes.json()
+      const sessionId = created.session.id as string
+      const { emitUserMessage, emitAssistantMessageStart, emitMessageDelta, emitMessageDone } =
+        await import('../src/server/events/session.js')
+
+      for (let turn = 1; turn <= 22; turn++) {
+        emitUserMessage(sessionId, `User ${turn}`)
+        const assistantId = emitAssistantMessageStart(sessionId)
+        emitMessageDelta(sessionId, assistantId, `Assistant ${turn}`)
+        emitMessageDone(sessionId, assistantId)
+      }
+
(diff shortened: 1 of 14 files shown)


locknet11:
I support this, in a large session history penalises browser performance and make it unusable.

JamesDAdams:
@S-Gr4y We made a big update, you need to refactor

PR #282, opened 2026-08-27
fix(llm): route Responses API models to /v1/responses

digitalsolutionscx:
## Problem

OpenCode Go serves a subset of its catalog through OpenAI's **Responses API** (`/v1/responses`) instead of `/v1/chat/completions` (per the official [endpoints table](https://opencode.ai/docs/go/#endpoints)):

| Model | ID | Endpoint |
|---|---|---|
| GPT 5.6 Luna | `gpt-5.6-luna` | `/v1/responses` |
| Grok 4.6 | `grok-4.6` | `/v1/responses` |
| Muse Spark 1.2 Contributor | `muse-spark-1.2-contributor` | `/v1/responses` |
| all others (GLM, Kimi, DeepSeek…) | … | `/v1/chat/completions` |

OpenFox's HTTP client was hardcoded to `{baseURL}/chat/completions`, so requests for these models were rejected by the Go gateway — only Luna/Grok/Muse crashed while every other model worked.

## Solution

**Per-model protocol routing** (not per-backend — one Go provider serves both endpoints):

- `responses-routing.ts` — declarative table mapping model IDs to their API protocol, matched case-insensitively on the full ID or last path segment, anchored to avoid prefix collisions (`grok-4.6-turbo` ≠ `grok-4.6`). Designed to host future protocols (Anthropic Messages).
- `responses-native.ts` — `OpenAIResponsesHttpClient` (same interface as `OpenAIHttpClient`/`OllamaHttpClient`) translating in both directions:
  - request: messages → `input[]` (system → top-level `instructions`), tools → flat `{type:'function',name,…}`, `max_tokens` → `max_output_tokens`, `reasoning_effort` → `reasoning:{effort}`, tool history → `function_call`/`function_call_output` items, `store:false` (stateless, ZDR-friendly)
  - response: `output[]` → content / `reasoning_content` / tool_calls, usage `input_tokens`/`output_tokens` → prompt/completion, status → finish_reason
  - stream: Responses SSE events (`response.output_text.delta`, `response.function_call_arguments.delta`, `response.completed`…) → standard `ChatCompletionChunk`s, so client.ts accumulation, idle timeout and abort handling work unchanged
- `client.ts` — `httpFor` now dispatches on the active model's protocol: ollama → native, Responses models → new client, default → chat/completions. Zero behavior change for existing models.

## Out of scope (follow-up PR)

Anthropic Messages models (`/v1/messages`: MiniMax M3/M2.7/M2.5, Qwen3.8/3.7/3.6 Max/Plus) still route to chat/completions and remain broken; the routing table is ready to receive them.

## Testing

- 21 new tests: routing unit tests, request/response/SSE translation unit tests (TDD), and end-to-end tests against a real `node:http` SSE mock asserting `/v1/responses` is hit, deltas accumulate correctly (content, tool calls, usage), and abort propagates
- `src/server/llm` suite: 223/223 pass — no regression
- typecheck + lint clean

### DIFF
--- a/src/server/llm/client-responses-e2e.test.ts
+++ b/src/server/llm/client-responses-e2e.test.ts
@@ -1,0 +1,203 @@
+import { afterAll, describe, expect, it } from 'vitest'
+import { createServer, type Server } from 'node:http'
+import type { AddressInfo } from 'node:net'
+import { createLLMClient } from './client.js'
+
+/**
+ * Boots a minimal SSE server emulating the OpenAI Responses API stream:
+ * text deltas, a function_call item, reasoning deltas, then response.completed.
+ * Records the request path so tests can assert /responses was hit.
+ */
+async function startResponsesMock(events: unknown[]): Promise<{
+  server: Server
+  port: number
+  requests: Array<{ path: string; body: Record<string, unknown> }>
+  abortControllerRef: { current?: AbortController }
+}> {
+  const requests: Array<{ path: string; body: Record<string, unknown> }> = []
+  const abortControllerRef: { current?: AbortController } = {}
+
+  const server = createServer((req, res) => {
+    let raw = ''
+    req.on('data', (chunk: Buffer) => {
+      raw += chunk.toString()
+    })
+    req.on('end', () => {
+      requests.push({ path: req.url ?? '', body: JSON.parse(raw || '{}') as Record<string, unknown> })
+      res.writeHead(200, { 'Content-Type': 'text/event-stream' })
+      for (const event of events) {
+        res.write(`data: ${JSON.stringify(event)}\n\n`)
+      }
+      res.end()
+    })
+    req.on('close', () => abortControllerRef.current?.abort())
+  })
+
+  await new Promise<void>((resolve) => server.listen(0, '127.0.0.1', resolve))
+  const { port } = server.address() as AddressInfo
+  return { server, port, requests, abortControllerRef }
+}
+
+const COMPLETED_EVENT = {
+  type: 'response.completed',
+  response_id: 'resp_1',
+  response: {
+    id: 'resp_1',
+    status: 'completed',
+    output: [],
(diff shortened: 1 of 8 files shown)


digitalsolutionscx:
## Functional testing (real Go subscription)

Validated end-to-end against `gpt-5.6-luna` on a live OpenCode Go account:

- **Simple prompt** → streams a response (previously rejected with HTTP 400 by the gateway)
- **Tool call** → `read_file` invoked with correctly accumulated JSON arguments (`{path:cahier-des-charges.md,offset:1,limit:5}`), tool executed, and the model cited the file content in its final answer — validating the `function_call`/`function_call_output` history translation and SSE argument-delta accumulation in real conditions
- **Multi-turn conversation** → context preserved across turns

Two fixes landed during testing, both driven by real gateway rejections:
1. `temperature` is not supported by GPT-5.x-family models on `/responses` → omitted (5c45e81a)
2. `top_p` rejected the same way → omitted (38d16fb1)

Only `max_output_tokens` is forwarded as a generation parameter now.

Automated: 21 new tests, `src/server/llm` suite 224/224, typecheck + lint clean. Non-Responses models (GLM, Kimi, DeepSeek…) are untouched by the routing change.

digitalsolutionscx (commented):
Bilan du test fonctionnel Luna :

Message simple ✓
Réponse en streaming ✓
Tool call avec arguments ✓
Suivi de conversation multi-tours ✓

digitalsolutionscx (commented):
Message simple ✓
Réponse en streaming ✓
Tool call avec arguments ✓
Suivi de conversation multi-tours ✓

co-l:
---
### AI-summary
- Reviewed the routing + Responses API client end to end: request/response/SSE translation, stateless store:false, and the per-model dispatch — no regressions (224 llm tests green, typecheck and lint clean)
- Verified empirically against a live Responses API that the tool+reasoning-effort combination works (the exact case chat/completions rejects for gpt-5-class models)

PR #285, opened 2026-08-28
fix(llm): apply OpenCode Go Responses routing on unknown backends too

digitalsolutionscx:
### Summary

Apply the OpenCode Go Responses-API routing table on `backend === 'unknown'` too.

The backend-aware refactor (c104613c) scoped the curated table to `opencode-go`, but providers created before it — or added with the backend left at its default — persist `backend: 'unknown'`. Their Responses-only models (`gpt-5.6-luna`, `grok-4.6`, `muse-spark-1.2-contributor`) then silently fell back to `/v1/chat/completions` and were rejected by the gateway (HTTP 400). Reproduced on a live Go subscription: same config, only `backend` differs → working / broken.

The curated ids are cloud-only model names that vLLM / SGLang / Ollama / llama.cpp / LM Studio never serve, so the fallback cannot mis-route a local model. `openai`, `anthropic` and explicit per-model overrides keep their precedence.

TDD: failing test `routes curated OpenCode Go models to responses on an unknown backend` first, then green.

### AI-Enhanced Development

- **AI Models:** GLM 5.3 Flash and GPT 5.6 Luna (via OpenFox agent sessions, OpenCode Go)

### Cache Impact

- **No** — routing decision only; request payloads and prompts are unchanged.

## Testing

- `src/server/llm`: 240/240 pass, typecheck + lint + prettier clean
- Live OpenCode Go subscription: `createLLMClient` with `backend: 'opencode-go'` reports `usesResponsesApi: true` and Luna answers with correct usage; the fix restores this without requiring the config change

### DIFF
--- a/src/server/llm/responses-routing.test.ts
+++ b/src/server/llm/responses-routing.test.ts
@@ -15,6 +15,15 @@ describe('resolveApiProtocol', () => {
     expect(resolveApiProtocol({ model: 'muse-spark-1.2-contributor', backend: 'opencode-go' })).toBe('responses')
   })
 
+  it('routes curated OpenCode Go models to responses on an unknown backend', () => {
+    // Providers created before the curated table existed (or added as "Other")
+    // carry backend 'unknown'; none of the local inference engines serve these
+    // ids, so the curated match still applies.
+    expect(resolveApiProtocol({ model: 'gpt-5.6-luna', backend: 'unknown' })).toBe('responses')
+    expect(resolveApiProtocol({ model: 'grok-4.6', backend: 'unknown' })).toBe('responses')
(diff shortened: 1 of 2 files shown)


digitalsolutionscx:
ready to go

digitalsolutionscx (commented):
fix(llm) with opencode-go ready to go

co-l:
Merged, will be released in v2.0.132
---
### AI-summary
- Reviewed the change: confirmed the `unknown` backend premise (providers created before the backend-aware refactor default to it) and that the anchored curated patterns cannot mis-route local inference engines
- Verified protocol precedence (explicit override > openai+profile > curated table > chat-completions) is preserved, with no test contradictions; full CI suite green
- Squash-merged onto develop

PR #295, opened 2026-08-30
fix(path-security): flag absolute paths nested in quotes

TheCause:
### Summary

An absolute path wrapped in single quotes nested inside double quotes escaped the path-security scan entirely — e.g. `python3 -c "print(open('/etc/passwd').read())"` or `bash -c "cat '/etc/passwd'"`, for reads and writes alike. The quote-pairing pass matched quotes without distinguishing their type, so the path became a boundary between two pseudo-strings and was never examined. The unquoted form was already caught, which masked the gap — and after `checkPathAccess` there is no downstream net in `run_command`.

The fix captures the opening quote in the absolute-path pattern and applies to nested spans the same regex-address rule the quoted-string pass already uses.

Proof by removal: reverting the fix turns 4 unit tests and 1 e2e test red. Locally green on this branch: lint, typecheck, duplicate, unit suite, e2e path-security (13/13). Originally written against 2.0.120, rebased onto current `develop` — the `maskCommandForPathScan` refactor is preserved.

### AI-Enhanced Development

- **AI Models:** Claude Fable 5 (via Claude Code) — code and tests, human-reviewed and human-tested before submission.

### Cache Impact

- **No**

https://claude.ai/code/session_01HPjMiPd4SeKQL8mKTw2XZ2

### DIFF
--- a/e2e/path-security.test.ts
+++ b/e2e/path-security.test.ts
@@ -118,6 +118,32 @@ describe('Path Security', () => {
       const payload = confirmationEvent!.payload as PathConfirmationPayload
       expect(payload.paths.some((p: string) => p.includes('home'))).toBe(true)
     })
+
+    it('emits path_confirmation for a path nested in quotes inside run_command', async () => {
+      client.clearEvents()
+
+      // The path lives in single quotes nested inside a double-quoted program
+      // string. Without quote-aware extraction no path is detected at all and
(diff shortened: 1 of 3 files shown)


co-l:
---
### AI-summary
- Reviewed the nested-quote path extraction fix: unit suite (282) and e2e path-security (13) green, typecheck and lint clean
- Rebased the PR onto latest develop
- Squash-merged

PR #296, opened 2026-08-30
fix(path-security): don't treat a search tool's PATTERN as a path

TheCause:
### Summary

In `grep [OPTIONS] PATTERN [FILE...]`, the first non-option operand is the PATTERN, never a file — but route-shaped patterns were extracted as absolute paths: `grep -rn "/api/routes" src/` asked for confirmation for a string that never touches the disk. In headless/API-driven sessions nobody can answer, so the agent stalls (measured: a 14-minute stall on a real run — the agent was grepping for an HTTP route that was the very subject of its task, so prompt-level warnings cannot prevent it).

This adds a masking pass for search-tool PATTERN operands (`grep`/`rg`/`egrep`/…) inside `maskCommandForPathScan`, before the regex-address pass. It follows the direction of existing tests that already expect regex patterns not to be treated as paths — route-shaped patterns like `/health` or `/api/x` only escaped confirmation by accident (a single letter after the closing slash happens to look like a sed address action).

Locally green on this branch: lint, typecheck, duplicate, unit suite, e2e path-security (12/12). Independent of #295 — either can merge alone.

### AI-Enhanced Development

- **AI Models:** Claude Fable 5 (via Claude Code) — code and tests, human-reviewed and human-tested before submission.

### Cache Impact

- **No**

https://claude.ai/code/session_01HPjMiPd4SeKQL8mKTw2XZ2

### DIFF
--- a/src/server/tools/path-security.test.ts
+++ b/src/server/tools/path-security.test.ts
@@ -676,6 +676,68 @@ describe('path-security', () => {
       })
     })
 
+    describe('search-tool patterns', () => {
+      // grep [OPTIONS] PATTERN [FILE...] — the first non-option operand is the
+      // pattern, never a file. A route-shaped pattern is the common case for an
+      // agent grepping its own codebase, and confirming it blocks headless runs.
+      it('does not extract a route-shaped grep pattern', () => {
+        const paths = extractAbsolutePathsFromCommand('grep -rn "/api/routage/gestes" portail/')
+        expect(paths).not.toContain('/api/routage/gestes')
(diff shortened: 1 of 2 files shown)


co-l:
---
### AI-summary
- Reviewed the search-tool PATTERN masking fix: unit suite (293) and e2e path-security (13) green, typecheck, lint and duplicate clean
- Rebased the PR onto latest develop (now including the merged #295) — both path-security PRs compose cleanly
- Squash-merged

PR #305, opened 2026-08-31
feat(commands): quoted arguments and {{ARGUMENTS}} for slash commands

nfertigsw:
### Summary

Command templates already took positional `{{name}}` placeholders, but the invocation was split on whitespace with no notion of quoting, and nothing could capture a free-form instruction. Both gaps send the user back to editing the command every time they want to vary it:

```
/revue src/a.ts "gestion des erreurs"   →  {{angle}} got `"gestion`, the rest was dropped
/note rerun the flaky proxy test        →  no placeholder could hold this
```

Two additions, both in a command's Message template:

- **Quoted values** — wrap an argument in `"…"` or `'…'` and it fills one placeholder whole. Inside double quotes a backslash escapes the next character; single quotes are literal, as in POSIX shells. Touching fragments join, so `src/"my file".ts` is one token. An unterminated quote takes the rest of the line rather than erroring — someone mid-keystroke has one most of the time, and the inline hint reads the same buffer.
- **`{{ARGUMENTS}}`** — receives everything typed after the command id, verbatim. Excluded from positional numbering so it never steals a token another placeholder was waiting for, and case-sensitive: `{{arguments}}` stays an ordinary slot.

Both flow through the existing unfilled-placeholder path, so a command invoked bare still opens the params modal instead of shipping `{{file}}` to the model.

Full design rationale and edge cases: **[docs/DESIGN-SLASH-ARGUMENTS.md](https://github.com/nfertigsw/openfox/blob/feat/slash-command-arguments/docs/DESIGN-SLASH-ARGUMENTS.md)** (added by this PR).

### One implementation instead of two

Parsing lived in two near-duplicates: `web/src/lib/parse-slash-command.ts` for the chat composer and `src/server/tasks/slash.ts` for commands seeded into task sessions — each with its own `extractTemplateParams` and its own `split(/\s+/)`. Both now delegate to `src/shared/slash-args.ts`, reachable from web through the existing `@shared` alias.

That was the point of the server-side copy existing in the first place: a command typed in chat and the same command seeded into a task should expand identically. They now do so by construction rather than by two implementations agreeing.

### Two surfaces that follow from the parser

- The composer's `param=?` hint counted arguments with `split(/\s+/)`, so a quoted value made it skip ahead. It now counts with the tokenizer, and `templateParamHints` orders `{{ARGUMENTS}}` last regardless of where it sits in the template, so the hint walks the slots in the order they have to be typed. Computed server-side in `routes/commands.ts`, so the composer and the task editor get identical hints from one place.
- The command editor gained a line describing the argument forms — including the `{{name}}` form, which nothing documented anywhere before.

### Backward compatibility

With no `{{ARGUMENTS}}` in the template, positional resolution is exactly what it was. With no quotes in the invocation, tokenization is exactly what it was. Every pre-existing test passes with only the two new fields (`args`, `rest`) added to the parse result.

Workflows share the tokenizer, so quoted values now work for workflow parameters too. `{{ARGUMENTS}}` does not apply to them — they declare typed parameters rather than a template.

### What to test

- **A multi-word value.** Give a command two placeholders, e.g. `Relis {{file}} en te concentrant sur {{angle}}.` Send `/cmd src/a.ts "gestion des erreurs"` and confirm the second placeholder gets all three words. Without the quotes it should still take just the first word, as before.
- **Free-form.** Write a command whose whole body is `Add a note: {{ARGUMENTS}}`, then send `/cmd rerun the flaky proxy test`. The full sentence should land in the prompt.
- **Mixed.** `Relis {{file}}. Consignes : {{ARGUMENTS}}` — the first token fills `{{file}}` and `{{ARGUMENTS}}` still carries the whole line, the same duplication `$1` + `$ARGUMENTS` produces elsewhere.
- **Nothing typed.** Send the same commands bare (`/cmd`) and confirm the params modal opens and asks, rather than the raw `{{file}}` reaching the model.
- **The inline hint.** Start typing `/cmd ` and watch the grey `name=?` hint. It should advance one slot per *token* — a quoted value counts once, not once per word — and offer `ARGUMENTS=?` only after the positional slots.
- **Nothing regressed.** Any existing command with plain positional arguments should behave exactly as before.
- **Tasks.** Seed a task whose prompt is a slash command with a quoted argument; the session should start with the expanded prompt, matching what chat produces.

### Tests

45 added. `src/shared/slash-args.test.ts` covers the tokenizer (quotes, escapes, touching fragments, unterminated quotes, empty quoted argument), parsing, hint ordering, and expansion. The two callers keep their own suites: `src/server/tasks/slash.test.ts` gains end-to-end `resolveSlashLaunch` cases for quoted arguments, `{{ARGUMENTS}}`, the mixed form, the bare-invocation refusal, and quoted workflow parameters; `ChatInput.slash.test.tsx` gains composer-level cases for a quoted value reaching the template and for an unfilled placeholder surviving to the modal.

<sub>Local run on WSL: `npm run typecheck` ✅, `eslint src/ web/src/` ✅ (0 issues), `npm run duplicate` ✅ (0 clones), `prettier --check` ✅, e2e ✅ (335 passed), unit 5000 passed with one failure — `inspect-proxy.test.ts > handles unreachable target gracefully`. That test proxies to `http://127.0.0.1:1` and expects a refused connection; on this machine the connect hangs instead (verified independently: a raw TCP connect to port 1 times out rather than being refused), so the proxy never answers and the test hits its 15s cap. It fails identically on a clean checkout without this branch. CI on ubuntu-latest should be unaffected.</sub>

### AI-Enhanced Development

- **AI Models:** Claude Opus 5 (1M context), via Claude Code

### Cache Impact

- **No.** Command templates are not part of the system prompt, the tool definitions, or the skill metadata that feed `computeDynamicContextHash` — an expanded command becomes an ordinary user message. This PR changes only how that message's text is assembled before it is sent, so nothing cached is invalidated or reshaped.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -198,6 +198,8 @@ Every file in `.openfox/` must be **committable** and **meaningful in the projec
 
 **Rationale:** Clean repo, no leaked personal config, reliable source of truth for CI and collaborators.
 
+**Command arguments:** a command's Message template takes `{{name}}` placeholders, filled positionally by order of appearance, and `{{ARGUMENTS}}`, which takes everything typed after the id. Quoted values (`/cmd "two words"`) fill one placeholder whole. Parsing lives in `src/shared/slash-args.ts` and is shared by the chat composer and the task board's server-side resolution — change it there, not in either caller. Full design: [docs/DESIGN-SLASH-ARGUMENTS.md](DESIGN-SLASH-ARGUMENTS.md).
+
 **Workflows:** When asked to create or edit a workflow, load the built-in `workflows` skill via `load_skill("workflows")` — it is the authoritative reference for workflow file format, storage locations (project `.openfox/workflows/` vs global `{configDir}/workflows/`), the full JSON schema, step types, transition conditions, and template variables ([docs/WORKFLOWS.md](docs/WORKFLOWS.md) is a pointer to it). Project workflows belong in `.openfox/workflows/` and are committable.
 
 ## TDD Workflow
--- a/docs/DESIGN-SLASH-ARGUMENTS.md
+++ b/docs/DESIGN-SLASH-ARGUMENTS.md
@@ -1,0 +1,87 @@
+# Slash Command Arguments
+
+## Goal
+
+Make command arguments usable for the two things people actually type: a value that contains spaces, and a free-form instruction.
+
+Commands already accepted positional arguments — `{{name}}` placeholders filled by order of appearance — but the invocation was split on `/\s+/` with no notion of quoting, and there was no way to capture a whole sentence. Both gaps push users back to editing the command every time they want to vary it.
+
+```
+/revue src/a.ts "gestion des erreurs"     → {{angle}} got `"gestion`, the rest was dropped
+/note rerun the flaky proxy test          → no placeholder could hold this
+```
+
+## User Experience
+
+Two additions, both in the command's Message template.
+
+**Quoted values.** Wrap an argument in `"…"` or `'…'` and it fills one placeholder whole.
+
+```markdown
+Relis {{file}} en te concentrant sur {{angle}}.
+```
+
+`/revue src/a.ts "gestion des erreurs"` → _Relis src/a.ts en te concentrant sur gestion des erreurs._
+
+**`{{ARGUMENTS}}`.** Receives everything typed after the command id, verbatim.
+
+```markdown
+Add a note to the plan: {{ARGUMENTS}}
+```
+
+`/note rerun the flaky proxy test` → _Add a note to the plan: rerun the flaky proxy test_
+
+Both forms coexist with what already worked: unfilled placeholders still open the params modal, so `/revue` alone asks for `file` and `angle` in a dialog. The command editor now carries a one-line description of all of this, which nothing documented before.
+
+## Technical Design
+
+### One module, both callers
+
+Parsing lived in two places that had drifted into near-duplicates: `web/src/lib/parse-slash-command.ts` for the chat composer and `src/server/tasks/slash.ts` for slash commands seeded into task sessions. Each had its own `extractTemplateParams` and its own `split(/\s+/)`.
+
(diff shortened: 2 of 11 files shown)


nfertigsw:
Hi @co-l — gentle nudge on this one.

Still `MERGEABLE` against `develop`, no conflicts, despite the 51 commits that have landed since I opened it. Happy to rebase onto 2.0.140 if you'd prefer it fresh.

One thing the diffstat hides: of the 612 additions, 321 are tests and 89 are docs. The production change is **+202 / −90 across six files**, and most of that is a single new module — `src/shared/slash-args.ts` — that replaces the two near-duplicate parsers in `web/src/lib/parse-slash-command.ts` and `src/server/tasks/slash.ts`. Those two are why a command typed in chat and the same command seeded into a task could drift apart; they now expand identically by construction. `npm run duplicate` reports 0 clones after the change.

If the size is what's giving you pause, I'm happy to split it — the tokenizer plus the dedup can land on its own, with `{{ARGUMENTS}}` following in a second PR.

JamesDAdams:
@nfertigsw Hi, can you resolve the conflicts, please?

nfertigsw:
@JamesDAdams Done — rebased onto `develop` (`09c271ef`), branch is green again.

Only one conflict, in `web/src/components/plan/ChatInput.tsx`: `develop` restructured the whole composer (indentation, `onFocus`/`onBlur`, `expandedHeight`, `QueuedMessages` moved above the `PluginSlot`, split mobile/desktop button rows). I took `develop`'s version of that block wholesale and re-applied the one line this PR changes there — the `filledCount` computation now goes through `tokenizeArgs` so a quoted run counts as a single argument. The other two hunks (the `expandCommandPrompt` import and its call in `onSendCommand`) merged cleanly.

No behaviour change from the rebase, and no new commits — still a single commit.

Verified on the rebased branch: `npm run typecheck`, `npm run lint`, `npm run format`, `npm run duplicate` (0 clones), `npm run build` and `npm run test:unit` (5599 passed). The only red test locally is `src/server/dev-server/inspect-proxy.test.ts > handles unreachable target gracefully`, which fails the same way on a clean `develop` checkout in my environment — it's network-sandbox related, not from this PR.

The offer from my previous comment still stands if you'd rather split it: the tokenizer + the server/web parser dedup can land on their own, with `{{ARGUMENTS}}` following in a second PR.

PR #314, opened 2026-09-03
fix mobile interface

DavidBabel:
### Summary

This PR is mainly cosmetic and fix toubles for mobile usage.
Since openfox is web based, it's a very good tool to use "on the move". And I really like it.

But even if the tool follow good convension regarding the mobile view, it sometimes miss an UX point of view.

This PR is here make that better.

## Fixes

**- Fix :** It's not possible anymore to zoom in with two fingers in mobile browser.
Why ? : because this is also enabled on input and textarea click with a default behavior on zooming on the field on focus.
On responsive interfaces, it is good practice to deactivate with `maximum-scale=1.0, user-scalable=no`

## Improvements

- Desktop mode is ISO with existing

PR: 

26-09-03-214258-ZsDQ

Upstream : 

26-09-03-214304-OeDi

- Improved view on 2 lines instead of a compressed one 
- Replace button text by an Icon in mobile mode to send a message (saves a lot of horizontal space)

PR :

image

Upstream : 

image

- Nice display of models modal

PR / Upstream

image  chrome_xxRILHUIvb

- Nice display of MCP modal

PR / Upstream

image image

- Full size text edition on phone when keyboard is open : 

PR / Upstream

IMG_2598 IMG_2599 

### AI-Enhanced Development

- AI Models: Qwen 3.8 Flash Next 110b Q4XL

### Cache Impact

- **No**


co-l:
---
### AI-summary
- Rebased the PR onto latest develop and resolved the merge conflict in `ChatInput` (kept both sides' changes — develop's pause/stop button rework + the PR's mobile layout)
- Refined the mobile composer: replaced the floating round send/stop buttons with the desktop-style pill buttons (icon-only on mobile, same height as the MoreMenu trigger), extracted shared `sendButton`/`stopButton` components
- Reworked the mobile footer into two balanced rows (agent + danger level on top, MCP + model below) and fixed the MCP dropdown overflowing the viewport by left-aligning the anchored panel on narrow containers
- Verified the full suite: unit tests green, typecheck + lint + duplicate checks clean
- Pushed the rebased branch to the contributor's fork so the PR merges cleanly with full attribution

0x3dlux:
Sorry, but can we please revert this "Full size text edition on phone when keyboard is open" part? It's driving me nuts! While typing, I want to see what I'm replying to. Especially when the LLM wants to have several questions answered at once, it is very useful to be able to scroll through the LLM's messages without having to jump in and out of a "modal" textarea. With your fix, the whole interface is redrawing and jumping around, twice per "what was the LLM's next question again?", each time the keyboard flies out, then comes back in while the interface itself does it's own set of animations... this is just disorienting and costs way too much mental context switching!

Ideally, the text area should occupy as much space as it's text contents needs, while still maintaining a scrollable see-through of at least 2~3 lines into the LLM's previous messages - exactly like it was before this fix. Putting the user in front of an empty, fullscreen textarea is like "leaving the app", dropping into a blank canvas, all state lost, not knowing what it was we were doing. Sorry, but I'm having a hard time figuring out what your actual goal was behind this. Just because someone declared this some sort of "best practice" back in 2012? It is the complete opposite for this use-case.

Also you seem to have introduced an actual bug with this: you have to press the "Send" button twice because it doesn't work right away in the maximised view: the first press of send "minimizes" the textarea, then only the 2nd press actually sends.

co-l:
Agreed with both feedbacks ; noticed both the full screen message and the double click for some reason.

@DavidBabel I'll fix the double click bug and move the full screen textarea into an opt-in option in the settings, so you'll still be able to select it

co-l:
Follow-up on both points — shipped in **v2.0.142**:

- **Fullscreen textarea is now opt-in** — a "Composer" toggle under Settings → Display, off by default. Default behavior is back to the auto-growing textarea with the conversation visible while typing.
- **Double-tap send fixed** — while the composer is expanded, pressing send/pause/stop (or the more menu) no longer collapses the layout mid-tap, so everything works on the first press.

Thanks again for the report!

DavidBabel:
Make sense, thanks for the feedbacks, I appreciate.

PR #325, opened 2026-09-07
fix: support Bearer token authentication for Ollama behind OpenWebUI

pledou:
### Summary

This allows Ollama to work with authentication providers like OpenWebUI that require Bearer token authentication.

- Use optional apiKey parameter to OllamaClientOptions
- Pass Authorization header with Bearer token when apiKey is provided
- Pass apiKey to OllamaHttpClient constructor in createLLMClient
- Enables usage with OpenWebUI proxy (https://openwebui.domain/ollama)

### AI-Enhanced Development

- Claude Haiku4.5

### Cache Impact

- **No**

### DIFF
--- a/src/server/llm/client.ts
+++ b/src/server/llm/client.ts
@@ -73,6 +73,7 @@ export function createLLMClient(
   // options.num_ctx). Dispatched per request based on the current backend.
   const ollamaHttpClient = new OllamaHttpClient({
     baseURL: stripVersionPrefix(baseURL),
+    apiKey: config.llm.apiKey,
   })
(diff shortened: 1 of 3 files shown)


co-l:
---
### AI-summary
- Fixed a typecheck failure (`exactOptionalPropertyTypes`): the `apiKey` was passed as `string | undefined` into optional props — now spread only when present, and the client field is explicitly `string | undefined`
- Added regression tests for the new header logic: `Authorization: Bearer` is sent when a key is configured, and absent when not (backward compatible with plain Ollama)
- Verified the full suite: server tests green, typecheck + lint + duplicate checks clean
- Pushed the updated branch to the contributor's fork so the PR merges cleanly with full attribution

PR #326, opened 2026-09-07
fix(llm): send x-opencode-session headers to opencode providers

ColasNahaboo:
## What

Since 2026-09-06, the Opencode Go & Zen providers require specific HTTP headers.

This PR makes OpenFox  send the headers OpenCode Go/Zen expects on inference requests to `opencode.ai` endpoints:

- `x-opencode-session: <session id>` — OpenFox's own stable session id, so Go routes same-session requests to the same provider and prompt caching stays effective
- `x-opencode-client: openfox`
- `User-Agent: openfox` (Go asks clients to identify themselves rather than the default HTTP-library UA)

Headers are sent on both the chat-completions and the OpenAI Responses paths.

## Why

OpenCode Go requires a stable `x-opencode-session` per conversation and started erroring on requests without it as of 2026-09-05 (https://opencode.ai/docs/go/ — 'Send a stable session ID in x-opencode-session'; vendor notice echoed in https://github.com/vercel/ai/issues/20271). 

Other harnesses had also to fix it, e.g:
- pi https://github.com/earendil-works/pi/issues/4847
- openclaw https://download.plaud.ai/openclaw/openclaw/issues/137165

## How

- `LLMCompletionRequest` gains an optional `sessionId`; the shared HTTP layer merges extra request headers (single merge point for all clients)
- `createLLMClient` detects opencode.ai targets (https + hostname match) and injects the headers
- The agent loop (chat turns, warmup, compaction, sub-agents) and session title generation pass their session id

Non-opencode providers are untouched.

### AI-Enhanced Development

- **AI Models:** GLM-5.3-Flash (via opencode) — code and PR text, human-reviewed and human-tested before submission.

### Cache Impact

- No

### DIFF
--- a/src/server/chat/agent-loop.test.ts
+++ b/src/server/chat/agent-loop.test.ts
@@ -622,6 +622,46 @@ describe('maxTokens clamping', () => {
     expect(callArgs.modelSettings?.maxTokens).toBe(2952)
   })
 
+  it('passes the sessionId to streamLLMPure for opencode session affinity', async () => {
+    mockSessionManager = {
+      enterPauseGate: vi.fn().mockResolvedValue('released'),
+      requireSession: vi.fn().mockReturnValue({
+        workdir: '/test',
+        projectId: 'test-project',
+        executionState: null,
+        criteria: [],
+        isRunning: false,
+      }),
+      getEffectiveWorkdir: vi.fn().mockReturnValue('/test'),
+      getProjectWorkdir: vi.fn().mockReturnValue('/test'),
+      getContextState: vi.fn().mockReturnValue({
+        currentTokens: 0,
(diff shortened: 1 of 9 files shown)


co-l:
---
### AI-summary
- Verified the header injection flows through the single shared HTTP merge point (`http-shared.ts`), covering chat-completions, streaming, and the OpenAI Responses path for `opencode.ai` targets
- Confirmed the `sessionId` plumbing end-to-end (agent loop warmup + main turns, stream layer, session title generation)
- Reviewed the 6 tests covering complete/stream/responses paths plus the negative cases (no session id, non-opencode targets); no changes were needed
- Verified the full suite: tests green, typecheck + lint + duplicate checks clean

PR #327, opened 2026-09-08
fix(llm): disable reasoning echo for opencode.ai providers

ColasNahaboo:
## What

Providers whose URL host is `opencode.ai` (Zen / Go) no longer echo chain-of-thought back on assistant history messages: openfox resolves `sendReasoningInMessages: false` from the URL **before** the provider's persisted value.

## Why

OpenCode Go's gateway forwards requests raw to upstreams whose strict schemas reject reasoning echoes on assistant history. GLM-5.3 via opencode go fails every multi-turn thinking request with:

```
HTTP 400: invalid_request_error — Extra inputs are not permitted,
field: messages[3].reasoning
```

`glm-5.3-flash` only dodged it because it is non-thinking and has nothing to echo.

The provider edit modal stamps `sendReasoningInMessages: true` on new providers (its default), so the persisted value is the foot-gun. That's why the URL default **overrides** the persisted flag: existing configs are fixed with no user action and no re-run of auto-config. Hosts without a known default keep the explicit provider value unchanged.

Same rescue pattern as the existing `api.deepseek.com` → `thinkingField: reasoning_content` URL default.

## How

- `backend.ts`: `UrlProviderDefaults` gains `sendReasoningInMessages`; `HOST_PROVIDER_DEFAULTS['opencode.ai'] = { sendReasoningInMessages: false }`
- `provider-manager.ts`: `resolveSendReasoningInMessages` (URL default first, explicit provider value as fallback) wired into `createConfigForProvider`
- Tests: URL default detection + 3 resolution cases (persisted `true` rescued → `false`; unset → `false`; non-opencode host keeps explicit value)

### AI-Enhanced Development

- **AI Models:** GLM-5.3-Flash (via opencode) — code and PR text, human-reviewed and human-tested before submission.

### Cache Impact

- System prompts / tool definitions / skills / openfox-internal context caching: No.
- Provider-side prompt caching on opencode.ai endpoints: Yes, one-time — assistant history no longer carries the `reasoning` echo, so in-flight conversations will cache-miss once after updating; the request shape is stable going forward (and multi-turn GLM-5.3 requests stop 400ing).

## Deliberately skipped

Changing the modal's blanket `true` default: the URL override already neutralizes it for opencode.ai, and changing modal semantics touches the onboarding flow and i18n strings. Can follow up if maintainers prefer the modal default to be 'auto' (unset).

### DIFF
--- a/src/server/llm/backend.test.ts
+++ b/src/server/llm/backend.test.ts
@@ -37,6 +37,11 @@ describe('backend', () => {
       })
     })
 
+    it('disables reasoning echo for the opencode.ai host', () => {
+      expect(detectProviderDefaultsFromUrl('https://opencode.ai/zen/go/v1')).toEqual({ sendReasoningInMessages: false })
+      expect(detectProviderDefaultsFromUrl('https://opencode.ai/zen/v1')).toEqual({ sendReasoningInMessages: false })
+    })
+
     it('returns undefined for local or unknown hosts', () => {
       expect(detectProviderDefaultsFromUrl('http://localhost:8000')).toBeUndefined()
       expect(detectProviderDefaultsFromUrl('http://192.168.1.223:8000/v1')).toBeUndefined()
--- a/src/server/llm/backend.ts
+++ b/src/server/llm/backend.ts
@@ -123,12 +123,19 @@ export function detectBackendFromUrl(url: string): Backend | undefined {
 export interface UrlProviderDefaults {
   /** Field the provider reads chain-of-thought from in assistant history. */
   thinkingField?: string
+  /** Whether to echo chain-of-thought back on assistant history messages. */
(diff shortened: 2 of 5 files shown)


co-l:
---
### AI-summary
- Verified the reasoning-echo fix: `opencode.ai` hosts now resolve `sendReasoningInMessages: false` from the URL, overriding a persisted `true` — same rescue pattern as the existing `api.deepseek.com` thinking-field default
- Confirmed the 3 resolution cases (persisted `true` rescued, unset, non-opencode host keeps explicit value) plus the URL detection test
- Noted the PR head also carries an unrelated commit (process-tree BFS cycle guard, test-only, Windows flake) — harmless, included as-is
- Verified the full suite: tests green, typecheck + lint + duplicate checks clean

PR #331, opened 2026-09-09
fix(sub-agents): resolve model overrides in workflow sub-agent steps

JamesDAdams:
### Summary

- Fix sub-agent model overrides resolution when executing sub-agents via workflows (`executor.ts`) and ensure fallback to `sessionManager.getProviderManager()` in `executeSubAgent`.
- Previously, `providerManager` was not passed to `executeSubAgent` during workflow sub-agent step execution, causing sub-agents to ignore their configured model overrides and silently fall back to the session's parent LLM client.

### AI-Enhanced Development

- **AI Models:** Gemini 3.7 Flash

### Cache Impact

- **No**

### DIFF
--- a/src/server/sub-agents/manager.test.ts
+++ b/src/server/sub-agents/manager.test.ts
@@ -335,6 +335,35 @@ describe('SubAgentManager', () => {
       expect(result.content).toBe('Test result content')
     })
 
+    it('falls back to sessionManager.getProviderManager() when providerManager is omitted in options', async () => {
+      resolveLLMClientForAgentMock.mockReturnValue({
+        client: createMockLLMClient(),
+        usedOverride: true,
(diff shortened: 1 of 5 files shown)


co-l:
Merged, will be released in v2.0.143
---
### AI-summary
- Reviewed the PR and confirmed the root cause: the workflow `sub_agent` step was the only `executeSubAgent` caller omitting `providerManager`, so per-sub-agent model overrides were silently ignored there
- Verified edge cases (explicit `providerManager` wins, session fallback when absent) and ran the full suite — 5266 unit tests green, typecheck and lint clean; flagged a pre-existing unrelated flaky e2e (`path-security`) that passes in isolation
- Squash-merged onto `develop` with full attribution

Issue #185, opened 2026-07-31
Do not automatically create a Git repo in the "project" directory


matthieu-labas:
I have a mono-repo with different sub-projects (`~/dev/code/the-project/project-*/` and `~/dev/code/the-project/.git/`).
I usually open some projects in OpenFox, and it seems it automatically create a Git repo in said project, which messes up my other tools.

To detect if a project has a Git repo, you can run `git rev-parse --show-toplevel` instead of `ls -d .git` (or whichever command is used).

Defragmentor:
I second this request.
Not everyone uses git. Some don't even use OpenFox for development ;-)

My suggestion is for a button to on/off this feature. Preferably on a per project base.

matthieu-labas:
I resorted to pulling the project and Qwen spent 6 min telling me this.

I'm no TS dev but it seems it could be implemented quickly. If you want I can make a PR but I'll have Qwen do it; probably not what you want...



===== Qwen analysis

## Root cause

The only place git is initialized is `createProjectDirectory()` in `src/server/utils/project-creator.ts:96-97`:

```ts
if (!(await directoryExists(join(fullPath, '.git')))) {
  execSync('git init', { cwd: fullPath, ... })
```

It runs unconditionally whenever a project is opened/registered, if the target dir has no .git. Two entry points call it, both with no way to opt out:
* Web UI — `CreateProjectModal` → `POST /api/projects` (`src/server/index.ts:407`)
* MCP — `openfox_create_project` → `deps.createProject` (`src/server/index.ts:3682`)

The check is also naive — it only looks for `.git` in the directory itself (no walk up), and on permission failure it even retries `git init` via `sudo -u $USER`. Everything else (the workspace tool, workspaces) correctly requires an existing repo rather than creating one.

## Plan

Add an `autoGitInit` config option, defaulting to `true` (current behavior preserved):

* Global config (`~/.config/openfox[-dev]/config.json`): `workspace.autoGitInit` (boolean, optional; `false` = opt out)
* Env override: `OPENFOX_AUTO_GIT_INIT` (env > global config > default `true`)
* Behavior: when `false`, `createProjectDirectory` skips the git block entirely (no `git init`, no sudo fallback) — still creates the directory and registers the project
* MCP tool description updated to reflect the conditional behavior

matthieu-labas:
GitHub is making that too painful. Changes are there: ~5 lines modified (+ tests).
Trying to target PR into develop branch but it keeps getting stuck on master...

[...]

PR #336, opened 2026-09-10
Option to skip git init on non-git projects.

matthieu-labas:
### Summary

Attempted correction for #185 
/!\ TESTS WERE NOT RUN (I don't write TS and don't have the testing frameworks) but modifications are so minimal that it should not break anything (or would probably be caught by CI)

### AI-Enhanced Development

- **AI Models:** 

Qwen3.8-27b on OpenFox 2.0.143

### Cache Impact

- **No**


matthieu-labas:
Redoing on branch develop...

PR #341, opened 2026-09-10
Feat/workflow step model override, new worlflow parameters type, zoom, move bloc in ui

JamesDAdams:
### Summary

Workflow edits now take effect immediately instead of only after a full page reload (F5).

Editing a workflow (parameters, defaults, options, steps) left every other surface — the composer's workflow picker, the launch parameter modal, slash-command hints, quick actions — serving the pre-edit definition until the page was reloaded. The workflow list (which carries parameter definitions) is cached client-side per workdir with a 60s freshness window, and a save only revalidated the editing modal's own scope: the composer's pencil path passed no scope at all, so the composer kept reading a different cache key.

Changes:
- **Cache primitives** (`web/src/lib/resourceCache.ts`): entries now remember their fetcher; `refreshMatching(prefix, includeKey?)` revalidates every live entry under a key prefix (single-flight, caller's key deduped), and `loadIfStale(key, fetcher, maxAgeMs)` refreshes only when an entry is older than a window.
- **Convergence after a save** (`web/src/lib/resources.ts`, `web/src/lib/workflows-actions.ts`): `refreshWorkflowLists(workdir)` runs after every create/update/delete/duplicate and revalidates the caller's scope plus every list a mounted consumer holds — user-scope edits are global, so all projects converge with one request per live list.
- **Interactive revalidation** (`revalidateWorkflows`, exposed as `useWorkflows().revalidate`): 2s freshness-gated, so pickers, slash hints and launches pick up out-of-band edits (file on disk, another tab) without a request storm.
- **Wiring**: `MoreMenu` passes `projectDir` to `WorkflowsModal` (the missing scope) and revalidates when its Workflows tab is shown; `PlanPanel` revalidates before deciding whether to open the parameter modal (also fixes a stale `sessionWorkdir` closure); `WorkflowsModal` branches copy-of-default vs in-place on the freshly fetched list; `ChatInput` and `QuickActionModal` revalidate on slash start / open.

Also on this branch:
- `fix(update)`: parse the installed version from localized CLI output (`Mis à jour : X`), which previously left the auto-restart confirmation timing out on non-English locales.
- `feat(workflows)`: per agent/sub-agent step model + provider + reasoning-effort override, parameter types (input, textarea, checkbox, select) with defaults/options, and zoom/pan/drag repositioning of workflow blocks with step positions persisted in the workflow file.

Tests: new `workflows-actions.test.ts` and `MoreMenu.test.tsx`, extended `resourceCache.test.ts`, `resources-workflows.test.ts` and `ChatInput.slash.test.tsx`. Gates run: typecheck, lint, format, unit (5435 passed) and e2e (346 passed). The 3 remaining unit failures (`src/cli/mcp.test.ts`, `src/server/auth.test.ts` RSA PKCS#1 verification) are pre-existing on Node 24, unrelated to this change and reproducible on a clean tree.

### AI-Enhanced Development
Models that helped shape this PR:
- **AI Models:** DeepSeek V4 Flash, DeepSeek Flash, GPT-5.6 Luna

### Cache Impact
Does the PR affect anything cached — system prompts, tool definitions, skills, other context?
- **Yes** — the client-side workflow-list resource cache is now revalidated on mutation and on user interaction (gated at 2s), so the UI converges without a reload. Nothing changes system prompts, tool definitions or skills, so the LLM prefix cache is untouched by the revalidation fix. Note that the step model/effort override on this branch can change the LLM request mid-run; the entry transition is gated behind an explicit Apply/Keep choice precisely so the prefix cache is not silently invalidated.


JamesDAdams:
@co-l I don't know if you agree with this pull request. I created it to add the ability to override the AI model we want to use in workflows, and I also needed the workflows to have parameters other than text inputs—like checkboxes and lists—and I added the ability to zoom in on the workflow to make it easier to analyze, as well as the ability to move blocks around because by default they’re pretty crammed together, and that can quickly become a nightmare to read.

PR #353, opened 2026-09-13
feat/ci: add pre-push hook to run PR verification checks

JamesDAdams:
### Summary

- Add a Husky `pre-push` hook (`.husky/pre-push`) to automatically run the full verification check suite (`lint`, `typecheck`, `format`, `duplicate`, `test`) in parallel with early-exit failure prior to pushing.
- Optimize PR CI pipeline (`.github/workflows/pr.yml`) by splitting the monolithic `test` job into parallel `test-unit` and `test-e2e` jobs.
- Accelerate CI E2E test execution by configuring `OPENFOX_E2E_MAX_WORKERS: 2` to leverage multi-core runners instead of sequential single-worker execution.

### AI-Enhanced Development

- **AI Models:** Gemini 3.7 Flash

### Cache Impact

- **No**


co-l:
Hey! Does it replace the precommit hook or is an addition to it?

I'm worried about having two back-to-back checks when I ask the agent to "commit & push"

JamesDAdams:
> Hey! Does it replace the precommit hook or is an addition to it?
> 
> I'm worried about having two back-to-back checks when I ask the agent to "commit & push"

Right now there's no hook; the hook I added simply runs the tests included in the PR locally on my machine to make sure everything is working properly before merging, so the PR goes through right away.

But I get that it's annoying because it means running the tests twice, once with the PR and once locally.

JamesDAdams:
If you don't like it i can change the hook to be optional and update the "commit and push" command to add in the prompt "is there are optional hook ask the user if he want to run the hook or not"

co-l:
Ok so that's for the PR workflow rather than the normal workflow.

Right now there is already a pre-commit hook, so with that change it would run twice: once on commit, and a second time on push.

The proper solution that fixes that is to remove the pre-commit hook and keep only the pre-push ; you can do it.

PR #360, opened 2026-09-14
feat(sessions): two-phase delete with configurable end-of-session routine

0x3dlux:
### Summary

Two-phase session delete: pressing **Delete session** no longer destroys history immediately. It starts a configurable *end-of-session* routine (default: the bundled `/cmd end-of-session` wrap-up command) and marks the session `closing`; the routine's last act deletes the session.

- **Delete session** (sidebar / in-chat footer) → `POST /api/sessions/:id/end-session`: queues the configured command and stamps `sessions.closing_at`.
- **Delete now** on a closing session → final delete straight away (confirmation required).
- **Cancel closing** → clears `closing_at` and, if the wrap-up prompt never got its turn (the session was busy), withdraws it from the queue so a kept session cannot wrap itself up later.
- **Skip → Close now** in the confirm dialog → plain delete path, no command runs.
- Bulk delete (delete-all / project teardown) never runs the routine.
- **Server restart mid-close** → boot re-queues the routine for every still-`closing` session (the queue is transient, `closing_at` is not), so the state never silently degrades.
- Sidebar shows a `closing` badge while the routine runs and flips it to a green `done` once it concluded; the in-chat footer reports "End-of-session command done" and offers **Delete session** / **Keep session**.

New setting `session.endOfSessionCommand` (Advanced settings tab), default `end-of-session`. An empty value disables the routine (immediate delete); a command that is configured but **cannot run now answers 409 and deletes nothing** rather than quietly falling back.

**Global instructions as a line database.** The routine keeps durable facts in OpenFox's own global instructions, so that store gets verbs on the endpoint the settings UI already uses — no new resource, no whole-value rewrite:

| Method | Path | Body | Effect |
| --- | --- | --- | --- |
| GET | `/api/settings/global_instructions` | — | read the value (unchanged) |
| PUT | `/api/settings/global_instructions` | `{value}` | replace the value (unchanged, used by the settings tab) |
| POST | `/api/settings/global_instructions` | `{line}` | append one line; an identical line already present is reported as `changed:false, matched:n` |
| PATCH | `/api/settings/global_instructions` | `{match,line}` | replace every whole-line exact match |
| DELETE | `/api/settings/global_instructions` | `{line}` | remove every exact-matching line |

Every write answers `{ key, changed, matched, lineCount, removed[], added[], message }` ("Appended 1 line (43 total)", "Replaced 2 occurrence(s)", "No line matched — GET the setting to list current lines"), so the caller gets evidence rather than a bare 200. Matching is whole-line only (no numbering, no substrings), blank/whitespace lines cannot be addressed, a multi-line or empty argument is a 400, a non-line-addressable key is a 405, and no match is a 404. Each edit is a read-modify-write of the one settings row, so headings, blank lines and markdown hard breaks around the target come back byte-identical.

**No hedging.** The dialog and the new availability hint under the setting derive availability from the same merged command list the server resolves against (project > user > defaults, plus the command's declared `paramNames`), so they state what will happen: routine → *Run / Close later* (stacked) / *Skip / Close now* (stacked); disabled → plain delete; missing or parameter-hungry command → the reason plus *Delete* only. The child shells of a session also learn the port the server actually listened on (`OPENFOX_PORT` / `OPENFOX_API`), so their tooling always targets the server that is actually running.

Migration (idempotent, applied on first server start, no data rewrite; `NULL` = not closing):

```sql
ALTER TABLE sessions ADD COLUMN closing_at TEXT;
```

`closingAt` travels in both the streamed session payloads and `session.state`, so streaming and page-reload render identically. Cancelling reconciles the client pane against the server timestamp (no client-clock drift).

### AI-Enhanced Development

- **AI Models:** Qwen3.8-FN (self-hosted via vLLM), driven by OpenFox (planner + builder + reviewer/verifier sub-agents)

### Cache Impact

- **No** — no system prompt, tool definition or skill content changed. Added: routes, a settings key, one DB column, a bundled default command file (invoked, never injected) and a startup env export for child processes.

### Verification

- `npm run typecheck`, `npm run lint`, `npm run format` clean; `npm run duplicate` → 0 clones.
- `npm test` fully green (unit 5451 passed / e2e 359 passed), also in the serial `CI=true` e2e mode the test job uses.
- New tests: 13 line-op unit tests, 15 e2e settings-verb tests, 14 end-of-session route/resolve tests, 6 availability-helper tests, 9 Sidebar dialog tests, 5 AdvancedTab hint tests, 13 store tests, 1 self-address e2e test, 3 port-guard tests.

**Test-flake hardening** (the first CI test-job failures traced to shared-host races, not product bugs): WHATWG fetch refuses a blocklist of ports that overlaps the OS ephemeral range, so a `listen(0)` test server could land on e.g. 6667 and every request to it failed with "bad port" depending on the ephemeral cursor. `test-port-guard.ts` (loaded via `setupFiles` by both vitest configs) retries ephemeral binds onto reachable ports; `e2e/ports.test.ts` pins its list against undici's own. Also fixed: per-tool waits in the danger-mode-switch batch, pty exit-log drain before teardown (was aborting runs via `EnvironmentTeardownError`), no fixed-port binds in the findFreePort exhaustion test, clamp-safe `..` margin in the /tmp-escape path test.


co-l:
Hello, can you elaborate on the use case of this? Are you deleting sessions to save space or cleaning up, and want a safer flow for deletion?

0x3dlux:
The main goal would be to distill the session's learnings/gotchas into the global instructions or per-project AGENTS.md. I call my custom /learn command pretty much after every session, but it's easy to lose track where this has been done and where not. Sorry for spamming the CI, Qwen has been going circles on this one pretty much unchecked - I tend to run around 50 OpenFox sessions in parallel these days at all times - not sure if healthy 🤣

PR #239, opened 2026-08-13
feat(permissions): server-side permission rules (ALLOW/DENY/ASK) with session-allow promotion

ikarys:
### Summary

Adds deterministic, server-side permission rules (ALLOW/DENY/ASK) for any tool action. Rules are evaluated in Node (not by the LLM), scoped globally or per-project, and merged at runtime with ephemeral session-scoped rules promoted from ASK via "Allow for this session".

**Rule engine** (`src/server/permissions/`)
- `schema.ts` — Zod validation: non-pattern tools (`web_fetch`, `call_sub_agent`, ...) restricted to DENY-only, no pattern
- `rules.ts` — `evaluateRulesWithMatch` returns matched rule; precedence DENY (3) > ALLOW (2) > ASK (1)
- `registry.ts` — load/save `permissions.json` (global: `~/.config/openfox/`, project: `.openfox/`)

**Enforcement** (`src/server/chat/`, `src/server/tools/`)
- `agent-loop.ts` — per-batch merge of disk rules + ephemeral session rules into `batchContext.permissionRules`
- `execute-tools.ts` — `evaluateToolGate` DENY-only gate for non-path tools (`web_fetch`, `call_sub_agent`, ...)
- `path-security.ts` — 3-phase evaluation across all targets (command + paths): DENY throws, ALLOW skips sandbox, ASK prompts. Sub-agents fail-closed on ASK. "Allow for this session" promotes ASK → ephemeral ALLOW
- Hardcoded safety guards: git `--no-verify` and dangerous commands always prompt, even with ALLOW rule

**Cleanup**
- `index.ts` — `clearAllowedPaths` on session delete + stop (purge ephemeral rules)
- Removed deprecated `path.confirm` WS message (REST only)
- `.openfox/permissions.json` gitignored (machine-specific, not committable)

**UI** (`web/src/components/settings/`, `web/src/components/shared/`)
- `permissions-shared.tsx` — `RuleForm`, `RuleRow`, `PermissionsList` with effect/pattern adapting to tool type
- `ProjectSettingsModal.tsx` — project-scoped Permission Rules section
- `PathConfirmationButtons.tsx` — "Allow for this session" button; hidden for `dangerous_command`/`git_no_verify`

**Tests** — 155 unit + 3 e2e (DENY blocks, ASK prompts, allow-for-session no re-prompt)

### AI-Enhanced Development

- **AI Models:** GLM-5.2, DeepSeek V4 Flash 0731

### Cache Impact

[...]

PR #363, opened 2026-09-16
feat: Force agents to use pnpm instead npm for security purpose

juanpa669:
## What
Adds a guard to the `run_command` tool that refuses `npm`/`npx` usage and returns the `pnpm` equivalent for security purpose. 

## Behavior
- `npm install` → rejected, suggests `pnpm install`
- `npx vite build` → rejected, suggests `pnpm dlx vite build`
- Pure detection (`detectNpmUsage`) + guard placed after the background-`&` check.

## Scope
Agent-side only. The OpenFox project itself stays npm-based (package-lock.json);
This only constrains the commands the agent runs. 

You can add inside your project folder a pnpm-workspace.yaml to configure `minimumReleaseAge` variables. 
This makes it possible to install packages that already have a few days to protect against supply chain problems. 

`minimumreleaseAge is a security feature introduced in pnpm v10.16.0 that prevents the immediate installation of new package versions, reducing the risk of supply chain poisoning attacks.`

```yaml 
# pnpm-workspace.yaml
minimumReleaseAge: 5760 # 4 days
minimumReleaseAgeStrict: true
minimumReleaseAgeIgnoreMissingTime: false

trustPolicy: no-downgrade
blockExoticSubdeps: true
trustLockfile: false

strictDepBuilds: true
dangerouslyAllowAllBuilds: false

audit:
  level: high
  ```
## Tests
- 11 new tests (9 `detectNpmUsage` cases + 2 guard integration cases)
- Full unit suite: 5414 passed
- tsc + lint clean

### AI-Enhanced Development

- **Qwen3.8-27B**

### Cache Impact

- **No**


JamesDAdams:
Hi @juanpa669,

Thanks for the contribution and the security focus!

However, hardcoding an unconditional block on `npm`/`npx` inside `src/server/tools/shell.ts` cannot be merged into core as-is for a few reasons:
1. **Global breaking change**: Many projects (including OpenFox itself, which relies on `package-lock.json` and `npm`) do not use `pnpm`. Blocking `npm` at the core shell level breaks standard workflows for all repositories.
2. **Project-level scope**: Enforcing package managers is typically a project-specific constraint rather than a universal server rule.

### Recommended alternatives to enforce `pnpm`:
1. **Via Project Instructions (`AGENTS.md`) or a Skill**:
   Adding a strict constraint in your project's `AGENTS.md` (e.g. *"Never use npm/npx. Always use pnpm and pnpm dlx."*) directs the agent effectively without modifying server code.
2. **Via Project Permissions Rules (coming in #239)**:
   The upcoming permission engine allows defining per-project rules in `.openfox/permissions.json`:
   ```json
   {
     "rules": [
       { "effect": "DENY", "tool": "run_command", "pattern": "npm *" },
       { "effect": "DENY", "tool": "run_command", "pattern": "npx *" }
     ]
   }

PR #369, opened 2026-09-19
feat(compaction): cumulative compaction digest of prior rounds

rankaiyx:
### Summary

Opt-in cumulative view of compaction history: at the start of context-window round N, optionally inject a labeled digest of the prior rounds' compaction summaries, so the LLM sees the full per-round history instead of only the single recursively-condensed seed summary.

- `context.digestRound` (env `OPENFOX_DIGEST_ROUND` or global config file): `0` = off (default, byte-identical to current behavior), `-1` = all prior rounds, `k` = most recent k
- The decision is stamped per compaction on `context.compacted` (mirrored in `turn.snapshot`, latest-wins) → deterministic replay, per-round audit log, survives `cleanupOldEvents` purge and restart (also closes the latent `contextWindows` snapshot gap)
- The digest is a **pure projection** of persisted seed summaries (snapshot-path ≡ raw-event path, parity tested), emitted as an auto-prompt message (`type: compaction-digest`) into the new window before the seed relocation; sub-agent scopes excluded
- `COMPACTION_PROMPT` gains an anti-imitation guard (inherited by the rejected-tool-call retry path)
- Config softening: an invalid optional `context` block logs a warning and is ignored instead of silently resetting the whole config file
- UI: `compaction-digest` label (en/fr); docs: design doc, `SESSION-DEBUGGING.md` sync, README env table

Verified: TDD throughout; full unit + e2e suite, typecheck, lint, duplicate check all green; live boot check on openfox-dev (production config/DB untouched).

### AI-Enhanced Development

- **AI Models:** Qwen3.8-27B-MTP (local, via OpenFox)

### Cache Impact

- **Yes** — two context-injection changes, both compaction-scoped: (1) a new optional `compaction-digest` auto-prompt message is injected into the new context window before seed relocation, only when `digestRound` is non-zero; (2) `COMPACTION_PROMPT` (the compaction user prompt, including the rejected-tool-call retry variant) gained the anti-imitation guard paragraph. System prompts, tool definitions, and skills are untouched; with the default `digestRound = 0` the cached context stays byte-identical to before.


rankaiyx:
Hey author, could you take a look? I think this feature is pretty important. To deal with memory loss in multi-turn conversations, I currently have to ask the AI to manually maintain a record doc in every single session. 

Here's the design doc — I personally think it's elegantly designed:

[./docs/DESIGN-CUMULATIVE-COMPACTION-SUMMARIES.md](https://raw.githubusercontent.com/rankaiyx/openfox/7162befd8ff29295ff064886b5d210d21fe2d5cc/docs/DESIGN-CUMULATIVE-COMPACTION-SUMMARIES.md)

rankaiyx:
Rebased to the latest branch.

**Why this PR is safe to merge:**


**What it does, in one sentence:** when a long session gets compacted many times, the model gradually "forgets" early rounds because each summary is condensed from the previous one — this PR optionally shows the model a digest of all previous round summaries, so it remembers the whole journey, not just the last step.


**Why it's low-risk:**


- It's **off by default**. With the default setting, every session, snapshot, and cached prompt is byte-for-byte identical to today. Nothing changes until a user explicitly opts in.

- It **adds no new storage, schema, or API**. The digest is a *pure projection* — think of it as a table of contents generated on the fly from pages that already exist, not new content. It re-reads summaries the system has already saved, computes the digest at the moment of compaction, and stores nothing extra. Because it's derived rather than stored, there is nothing to migrate, nothing to keep in sync, and it can never go stale: same history → same digest, every time.

- It only kicks in at the exact moment of compaction, and never inside sub-agent scopes.


**Why it's elegant:**


- Instead of inventing a new mechanism, it **stamps one small number onto an event that already exists** and derives everything else by replay — the same event-sourcing pattern the codebase already uses, so behavior is deterministic across restarts and cleanups.

- The same digest renders identically whether the session is live or reloaded from a snapshot (parity is test-covered) — a direct consequence of the projection being stateless.


Happy to walk through the design doc (`docs/DESIGN-CUMULATIVE-COMPACTION-SUMMARIES.md`) or any specific hunk.

PR #233, opened 2026-08-09
feat(dev-server): opt-in Tailscale preview via dev.json config

theshwal:
## Summary

Adds an optional **Tailscale preview** to the dev server, enabled per project through a new `tailscaleExpose` boolean in `.openfox/dev.json` (default `false`).

The preview starts automatically after a successful dev server start, remains non-blocking on Tailscale errors, and is cleaned up on stop, crash, or stopAll — without touching pre-existing Serve entries.

## Feature

- **Opt-in via `tailscaleExpose` in `.openfox/dev.json`** — defaults to `false`; existing projects are unaffected.
- **New "Expose via Tailscale" checkbox** added to the Dev Server Config modal, under *Hot Reload* and *Disable inspect feedback*. The checkbox state is persisted with the rest of the dev-server config.
- **No-op on toggle while running** — the preview lifecycle is bound to dev-server Start/Restart, not to live toggling.
- **Auto-launch** on successful dev server start when `tailscaleExpose: true`.
- **Auto-cleanup** on Stop / crash / `stopAll` (no orphan entries).
- **Read-only Tailnet URL** in the footer when the preview is active (no separate Expose / Retry / Stop buttons in the UI).
- **Non-blocking on errors** — Tailscale failures (binary missing, daemon stopped, operator permission denied) are reported as `tailscalePreview.status: 'error'` and do **not** affect dev server state.

## Mechanism

- **`foreground` `tailscale serve` only** — entry disappears naturally with the controlling process. No `--bg` fallback.
- **`stdout` URL is primary** source for the Tailnet URL (matches Tailscale's actual exposed entry).
- **`tailscale serve status --json`** is used only to:
  1. Pick a free HTTPS port (skipping root `Web`/`TCP` *and* foreground-owned `Foreground[*].Web`/`TCP` entries).
  2. Confirm stabilization after spawn.
  3. Verify disappearance after kill.
- **Targeted cleanup** when an entry persists (`tailscale serve --https=<port> off`) — **never** `reset`. Pre-existing entries (e.g. 443 → 10369) are never touched.

## Tests

- **Unit (`src/server/dev-server/tailscale-preview.test.ts`)** — 25 tests covering foreground spawn, stdout URL parsing, `Foreground[*]` status JSON normalization (with a host-agnostic redacted fixture), targeted inverse, never-reset guarantee, stopAll.
[...]

PR #370, opened 2026-09-20
feat(plugins): expose dev-server lifecycle hooks

theshwal:
### Summary

Adds two generic Plugin API v2 lifecycle hooks for OpenFox dev servers:

* `devserver.started`
* `devserver.stopped`

This PR contains **no Tailscale-specific code**.

### Motivation

Plugin API v2 makes external integrations the right place for service-specific behavior, but plugins currently cannot observe the dev-server lifecycle.

A concrete consumer is the public community plugin:

https://github.com/theshwal/openfox-tailscale

That plugin keeps all Tailscale process management, settings, UI, status parsing, and cleanup outside OpenFox core. The only core capability it needs is a stable notification when a project dev server starts or stops.

This is the extension-point-only follow-up to the architectural feedback on #233.

#### `devserver.started`

Emitted after the dev server enters the running state.

```ts
{
  projectId?: string,
  data: {
    workdir: string,
    url: string,
    command: string,
    port: number
  }
}
```

#### `devserver.stopped`

Emitted for explicit stop, process exit, or process error.

```ts
{
  projectId?: string,
  data: {
    workdir: string,
    url: string | null,
    reason: 'stop' | 'exit' | 'error',
    exitCode?: number | null,
    error?: string
  }
}
```

A per-instance guard prevents duplicate stop events when an explicit stop also triggers the child-process close event.

### Project scope

When the dev-server workdir belongs to a known OpenFox project, the hook includes the normal top-level `projectId`. This allows plugins to use project-scoped settings without coupling core to a specific integration.

Dev-server lifecycle is workdir-scoped rather than session-scoped. The existing `PluginHookPayload` currently requires `sessionId`, so these events keep the existing contract and emit `sessionId: ''`. The canonical scope is `data.workdir`.

### Tests

Adds coverage for:

* resolved workdir / URL / command / port on `devserver.started`
* projectId propagation when a project can be resolved
* exactly one `devserver.stopped` event for an explicit stop

### Scope

This PR intentionally does not add:

* Tailscale dependencies or binaries
* dev.json settings
* core UI
* service-specific process management
* any plugin registry entry

Those belong to the external plugin.

### AI-Enhanced Development

* **AI Models:** GPT-5.6 Sol

### Cache Impact

* **No**


JamesDAdams:
@theshwal Nice, you can make a PR to add your plugin in the plugins-registry.json too

co-l:
Merged, will be released in v2.0.156
---
### AI-summary
- Rebased the PR onto latest develop and confirmed it merges cleanly (MERGEABLE/CLEAN, no conflicts)
- Ran the full verification suite during review: unit + e2e green, typecheck and lint clean
- Applied review fixes: added 3 regression tests covering the `devserver.stopped` close-path (clean exit, non-zero exit, and dedup when `close` fires after an explicit stop), and extracted a shared `buildHookPayload` helper to remove triple-duplicated payload construction
- Squash-merged the PR, then landed the review fixes as a follow-up commit on develop, re-verified green

